PR #4, opened 2018-10-19
MODSOURMAN-4 Created raml definition with stub implementation for Metadata Provider

KaterynaSenchenko:


descriptors/ModuleDescriptor-template.json
```diff
@@ -12,11 +12,11 @@
       "handlers": [
         {
           "methods": ["GET"],
-          "pathPattern": "/source-record-manager-ping"
+          "pathPattern": "/source-record-manager/logs"
```
Igor-Gorchakov: I would prefer to mention (in prefix) the exact component's name, which is responsible for handling incoming request:
for the Metadata Provider component let's use /metadata-provider/ prefix, the same for others

PR #6, opened 2018-11-09
Modsourman 7 MetadataRepository: Implement API for JobExecution entities

Igor-Gorchakov:
MetadataRepository: Implement API for JobExecution entities


ramls/jobExecution.json
```diff
@@ -11,12 +11,21 @@
       "description": "Human readable id",
       "type": "string"
     },
-    "jobProfileName": {
-      "description": "Name of applied job profile",
+    "parentJobId": {
+      "description": "",
```
KaterynaSenchenko: Please add description

OleksiiKuzminov:
As a suggestion. Maybe it will be a good idea to make a single architecture of modules services. Take a look on @KaterynaSenchenko PR and mod-source-record-storage

PR #9, opened 2018-11-19
Modsourman 9 module structure

Igor-Gorchakov:
- Module structure reorganization 
 (now module structure the same as for mod-data-import and mod-source-record-storage)
- Generic DAO interface
- Conversion JobExecution entity to DTO
- Unit tests
- Rest tests
- NEWS.md
- Snapshot version (Pom.xml)


pom.xml
```diff
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-source-record-manager</artifactId>
-  <version>0.0.1-SNAPSHOT</version>
+  <version>1.0.1-SNAPSHOT</version>
```
KaterynaSenchenko: Shouldn't it be 1.0.0-SNAPSHOT, same as in NEWS.md?

Igor-Gorchakov: You are right!

PR #10, opened 2018-11-19
MODSOURCE-14 Updated totalRecords to return overall total

KaterynaSenchenko:


ramls/examples/jobExecution.sample
```diff
@@ -1,6 +1,6 @@
 {
-  "jobExecutionId": "67dfac11-1caf-4470-9ad1-d533f6360bdd",
-  "jobExecutionHrId": "10000",
+  "id": "67dfac11-1caf-4470-9ad1-d533f6360bdd",
+  "hrId": "10000",
   "jobProfileId": "1",
```
Igor-Gorchakov: This is mine mistake, please change "jobProfileId" to the "jobProfileName" also, this is a temporal approach in order to read JobProfile name from the database. As soon as We get to know about JobProfile entity structure and storage, "jobProfileName" will be rolled back to "jobProfileId" again.

KaterynaSenchenko: I changed it in data-import-raml-storage, so no need to update here since it will be removed anyway.

KaterynaSenchenko: please verify that everything is correct in https://github.com/folio-org/data-import-raml-storage/pull/2

Igor-Gorchakov: perfecto

PR #25, opened 2018-12-03
Modsourman 11 - parse records and update them in storage

OleksiiKuzminov:


src/main/java/org/folio/services/parsers/SourceRecordParserBuilder.java
```diff
@@ -0,0 +1,25 @@
...
+  private SourceRecordParserBuilder() {
+  }
+
+  /**
+   * @param format - source record format
+   * @return - SourceRecordParser for chosen record format
+   */
+  public static SourceRecordParser buildParser(RecordFormat format) {
+    Optional<SourceRecordParser> sourceRecordParser = availableParsers.stream().filter(parser -> parser.getParserFormat().equals(format)).findFirst();
+    return sourceRecordParser.orElse(null);
```
TarasSpashchenko: null is not a good idea to return here, it is better to raise an informative exception or return a dummy/stub implementation which raises an exception when you try to call its methods

PR #41, opened 2018-12-20
Optimize code in source-record-manager

OleksiiKuzminov:


src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -251,45 +170,13 @@ private RecordFormat getRecordFormatByJobExecution(JobExecution execution) {
     Future<JobExecution> future = Future.future();
     RestUtil.doRequest(params, RECORD_SERVICE_URL + "/" + record.getString("id"), HttpMethod.PUT, record.encode())
```
KaterynaSenchenko: Maybe we should follow the same approach as with Snapshot entities - define Record type in raml instead of using JsonObject

src/main/java/org/folio/util/RestUtil.java
```diff
@@ -21,6 +27,8 @@
   public static final String OKAPI_TOKEN_HEADER = "x-okapi-token";
   public static final String OKAPI_URL_HEADER = "x-okapi-url";
   public static final int CREATED_STATUS_CODE = 201;
+  private static final String HTTP_ERROR_MESSAGE = "Response HTTP code not equals 200. Response code: ";
```
KaterynaSenchenko: *is not equal 200

src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -306,58 +305,26 @@ private Future saveSnapshots(List<JsonObject> snapshots, OkapiConnectionParams p
...
+        } else {
+          jobExecutionDao.updateBlocking(jobExecution.getId(), jobExec -> {
+            Future<JobExecution> jobExecutionFuture = Future.future();
+            jobExec.setErrorStatus(JobExecution.ErrorStatus.SNAPSHOT_UPDATE_ERROR);
+            jobExec.setStatus(JobExecution.Status.ERROR);
+            jobExec.setUiStatus(JobExecution.UiStatus.ERROR);
+            jobExecutionFuture.complete(jobExec);
+            return jobExecutionFuture;
+          }).setHandler(jobExecutionUpdate -> {
+            String message = "Couldn't update snapshot status for jobExecution with id " + jobExecution.getId();
```
Igor-Gorchakov: Can not update ...

src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -171,58 +172,56 @@ public JobExecutionServiceImpl(Vertx vertx, String tenantId) {
...
+        .withStatus(isSingle ? JobExecution.Status.NEW : JobExecution.Status.PARENT)
+        .withUiStatus(isSingle ?
+          JobExecution.UiStatus.valueOf(Status.valueOf(JobExecution.Status.NEW.value()).getUiStatus())
+          : JobExecution.UiStatus.valueOf(Status.valueOf(JobExecution.Status.PARENT.value()).getUiStatus()));
+    }
+    return job;
+  }
+
   /**
    * Creates and returns list of Snapshot entities (represented as JsonObject) depending on received JobExecution entities.
```
Igor-Gorchakov: please remove "(represented as JsonObject)"

src/main/java/org/folio/services/JobExecutionServiceImpl.java
KaterynaSenchenko: "Couldn't" is fine, but if you prefer using present simple, then "cannot" is much more common

PR #47, opened 2019-01-04
MODSOURMAN-48 Children JobExecution collection made searchable and pageable

KaterynaSenchenko:
* GET children by parent id made searchable and pageable
* default limit value set to some big number

### DIFF
--- a/ramls/change-manager.raml
+++ b/ramls/change-manager.raml
@@ -98,6 +98,20 @@ traits:
     /children:
       get:
         description: Get JobExecution children
+        is: [
+          searchable: {
+            description:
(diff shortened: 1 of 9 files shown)


ramls/change-manager.raml
```diff
@@ -98,6 +98,20 @@ traits:
...
+          searchable: {
+            description:
+              "with valid searchable fields: for example status=IMPORT_IN_PROGRESS",
+            example:
+              "status=IMPORT_IN_PROGRESS"
+          },
+          pageable,
+          validate
+        ]
+#       override limit default value in order to return all JobExecution children by default
```
OleksiiKuzminov: move to description

PR #67, opened 2019-03-01
MODSOURMAN-70 Created ChunkProcessingService

KaterynaSenchenko:


src/main/java/org/folio/services/ChunkProcessingService.java
```diff
@@ -0,0 +1,18 @@
...
+import io.vertx.core.Future;
+import org.folio.dataimport.util.OkapiConnectionParams;
+import org.folio.rest.jaxrs.model.RawRecordsDto;
+
+public interface ChunkProcessingService {
+
+  /**
+   * Process chunk of RawRecords
+   *
+   * @param chunk  - {@link RawRecordsDto} chunk with list of raw records
```
VictorSoroka1: Could you please align params like in the code below

https://github.com/folio-org/mod-source-record-manager/blob/modsourman-74/src/main/java/org/folio/services/ChangeEngineServiceImpl.java#L113

KaterynaSenchenko: It would make more sense to change the parameter order in the specified file, since in all our service interfaces okapi params are going the last

PR #74, opened 2019-03-11
MODSOURMAN-83 Reuse GenericAsyncAnswer in mod-source-record-manager

RuslanLavrov:


mod-source-record-manager-server/pom.xml
```diff
@@ -72,7 +72,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>data-import-utils</artifactId>
-      <version>1.1.0-SNAPSHOT</version>
+      <version>1.2.0-SNAPSHOT</version>
```
KaterynaSenchenko: dependency should be on the releases version - 1.1.0

PR #75, opened 2019-03-14
Modsourman 67 Remove JobProfile from the module 

OleksiiKuzminov:
* Add jobExecution dependency on DataType (MARC, BIBFRAME, EDIFACT, etc.)
* Implement the stub getRecordFormatByJobExecution method in ChangeEngineService based on the dataType associated with JobExecution
* Remove jobProfile schema from mod-source-record-manager schemas
* Change /change-manager/jobExecutions/id/jobProfile endpoint implementation that assigns JobProfile to JobExecution
* Update calls to the mentioned above endpoint from mod-data-import module

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -127,7 +127,7 @@ public class ChangeEngineServiceImpl implements ChangeEngineService {
   }
 
   /**
-   * STUB implementation until job profile is't exist
+   * Get records format for parser from Job Execution
    *
    * @param jobExecution - job execution object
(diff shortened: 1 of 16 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -130,8 +132,8 @@
    * @param jobExecution - job execution object
    * @return - Records format for jobExecution's records
    */
-  private RecordFormat getRecordFormatByJobExecution(JobExecution jobExecution) { //NOSONAR
-    return RecordFormat.MARC;
+  private RecordFormat getRecordFormatByJobExecution(JobExecution jobExecution) {
```
KaterynaSenchenko: Update the javadoc for this method

PR #85, opened 2019-04-15
MODSOURMAN-94 Parsed MARC Records are mapped to Instances and saved in mod-inventory

KaterynaSenchenko:
* Borrowed mapping of MARC to Instance logic from mod-data-loader. After parsing Records are mapped to Instances and saved in mod-inventory.

* Code placed in 
> org/folio/services/mappers/processor
 
**DO NOT need to be reviewed!** 
package is borrowed from mod-data-loader and will not be changed for now since it is a temporary solution.


mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -0,0 +1,113 @@
...
+   * @return the chunks
+   */
+  public static Iterator<String> runSplitFunction(String funcName, String val, String param){
+    if(val == null){
+      return null;
+    }
+    if(SPLIT_FUNCTION_SPLIT_EVERY.equals(funcName)){
+      return splitEvery(val, param);
+    }
+    return null;
```
Anatolii-Starkov:
```suggestion
    return nonNull(val) && SPLIT_FUNCTION_SPLIT_EVERY.equals(funcName) ? splitEvery(val, param) : null;
```

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -0,0 +1,113 @@
...
+  private NormalizationFunctions() {
+    throw new UnsupportedOperationException("Cannot instantiate utility class.");
+  }
+
+  /**
+   * Split val into chunks of param characters if funcName is "split_every".
+   * Return null if val is null or funcName is not "split_every".
+   * @return the chunks
+   */
+  public static Iterator<String> runSplitFunction(String funcName, String val, String param){
```
Anatolii-Starkov: Add parameter's descriptions

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -0,0 +1,113 @@
...
+      return splitEvery(val, param);
+    }
+    return null;
+  }
+
+  /**
+   * Run the function funcName on val and param.
+   * @return the function's result
+   */
+  public static String runFunction(String funcName, String val, String param){
```
Anatolii-Starkov: Add parameter's description

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -0,0 +1,113 @@
...
+        String []range = pos.split("-");
+        return val.substring(Integer.parseInt(range[0]), Integer.parseInt(range[1])+1);
+      }
+      int p = Integer.parseInt(pos);
+      return val.substring(p,p+1);
+    }
+    catch(Exception e){
+      return val;
+    }
+  }
```
Anatolii-Starkov:
```suggestion
    try{
      return pos.contains("-") ? handleDashCharacter(val, pos) : handleCharacter(val, pos);
    }
    catch(Exception e){
      log.debug("something went wrong ...");
      return val;
    }
```

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -0,0 +1,113 @@
...
+
+  private static String trimPeriod(final String input) {
+    if (input.endsWith(".")) {
+      return input.substring(0, input.length()-1);
+    }
+    return input;
```
Anatolii-Starkov:
```suggestion
 return input.endsWith(".") ? input.substring(0, input.length() - 1) : input;
```
return statement should be one

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -0,0 +1,113 @@
...
+        else {
+          //if ends with .. or . remove just one .
+          return input.substring(0, input.length()-1);
+        }
+      }
+      catch(IndexOutOfBoundsException ioob){
+        return input;
+      }
+    }
+    return input;
```
Anatolii-Starkov: RETURN should be one. Make the result var instead multiple returns

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -0,0 +1,113 @@
...
+      return input.substring(0, input.length()-1);
+    }
+    return input;
+  }
+
+  private static String removeEndingPunc(String val){
+    return modified(val, (val.length()-1));
+  }
+
+  private static String modified(final String input, int pos){
```
Anatolii-Starkov: JavaDoc + JavaExamples

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/ProcessedSingleItem.java
```diff
@@ -0,0 +1,28 @@
+package org.folio.services.mappers.processor;
+
+public class ProcessedSingleItem {
```
Anatolii-Starkov: JavaDocs

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/ProcessedSinglePlusConditionCheck.java
```diff
@@ -0,0 +1,19 @@
+package org.folio.services.mappers.processor;
+
+public class ProcessedSinglePlusConditionCheck extends ProcessedSingleItem {
```
Anatolii-Starkov: JavaDocs

KaterynaSenchenko: Thanks for your review, please keep in mind that code placed in org/folio/services/mappers/processor package is borrowed from mod-data-loader and will not be changed for now since it is a temporary solution.

Anatolii-Starkov: Thank goodness!

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/MarcToInstanceMapper.java
```diff
@@ -0,0 +1,20 @@
...
+import io.vertx.core.json.JsonObject;
+import org.folio.rest.jaxrs.model.Instance;
+import org.folio.services.mappers.processor.Processor;
+import org.folio.services.parsers.RecordFormat;
+
+public class MarcToInstanceMapper implements RecordToInstanceMapper {
+
+  @Override
+  public Instance mapRecord(JsonObject parsedRecord) {
+    return new Processor().process(parsedRecord);
```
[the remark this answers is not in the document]
KaterynaSenchenko: The mapping logic of the Processor is pretty weird, it saves elements of parsed instance into the object state making it unsafe to use the same object for multiple records. Hopefully, we'll get rid of that soon.

PR #90, opened 2019-05-02
MARC 999 field: Put Inventory instance UUID

Igor-Gorchakov:
* Development part - added blocking execution for post-processing steps of parsed records
* No tests provided


mod-source-record-manager-server/src/main/java/org/folio/services/ChunkProcessingServiceImpl.java
```diff
@@ -41,20 +68,60 @@
...
-            .map(result -> true);
-        }
+      .compose(parsedRecords -> {
+        jobExecutionSourceChunkDao.update(jobExecutionSourceChunk
+          .withState(JobExecutionSourceChunk.State.COMPLETED)
+          .withCompletedDate(new Date()), params.getTenantId());
+        checkIfProcessingCompleted(jobExecutionId, params.getTenantId())
+          .compose(completed -> {
+            if (completed) {
+              postProcessRecords(parsedRecords, jobExecutionSourceChunk, jobExecutionId, params);
```
KaterynaSenchenko: Please move postProcessRecords out of completed block as we discussed.

PR #91, opened 2019-05-03
MODDATAIMP-119 Added support for records coming for processing in json format

KaterynaSenchenko:


mod-source-record-manager-server/src/main/java/org/folio/services/parsers/JsonRecordParser.java
```diff
@@ -0,0 +1,35 @@
...
+        .put("error", rawRecord));
+    }
+    return result;
+  }
+
+  // Currently there is no corresponding RecordFormat for json since any type of Record can be stored in json format
+  // Implementation will be changed in scope of (@link https://issues.folio.org/browse/MODSOURMAN-112)
+  @Override
+  public RecordFormat getParserFormat() {
+    return null;
```
Igor-Gorchakov: Better throw UnsupportedOperationException

mod-source-record-manager-server/src/main/java/org/folio/services/parsers/RecordParserBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+   *
+   * Currently this method builds specific parser based on the record format
+   * Implementation will be changed in scope of (@link https://issues.folio.org/browse/MODSOURMAN-112)
+   *
+   * @param format - raw record format
+   * @param rawRecord - raw record itself
+   * @return - RecordParser for chosen record format
+   */
+  public static RecordParser buildParser(RecordFormat format, String rawRecord) {
+    if (rawRecord.startsWith("{")) {
```
Igor-Gorchakov: starts with { or [

KaterynaSenchenko: Nope, this is a record from RawRecordsDto, it contains separate records in whichever format it is. Here only single json can be in that String

PR #92, opened 2019-05-07
MODSOURMAN-108 - Send Records to SRS in bulk

RuslanLavrov:


OleksiiKuzminov:
rebase from master plz

PR #104, opened 2019-05-16
Modsourman 118 - 999 field: Instance UUID does not appear

Igor-Gorchakov:


ramls/instance.json
```diff
@@ -127,70 +195,169 @@
...
+      "type": "string",
+      "description": "Instance status term (e.g. cataloged, uncatalogued, batch loaded, temporary, other, not yet assigned)"
+    },
+    "statusUpdatedDate": {
+      "type": "string",
+      "description": "Date [or timestamp] for when the instance status was updated"
+    },
+    "metadata": {
+      "type": "object",
+      "$ref": "raml-util/schemas/metadata.schema",
```
KaterynaSenchenko: raml-storage/raml-util/schemas/metadata.schema

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/InstanceProcessingServiceImpl.java
```diff
@@ -45,36 +45,40 @@ public InstanceProcessingServiceImpl(@Autowired JobExecutionSourceChunkDao jobEx
...
+      if (result.failed()) {
+        LOGGER.error("Couldn't create Instance in mod-inventory", result.cause());
+        jobExecutionSourceChunkDao.getById(sourceChunkId, params.getTenantId())
+          .compose(optional -> optional
+            .map(sourceChunk -> jobExecutionSourceChunkDao.update(sourceChunk.withState(JobExecutionSourceChunk.State.ERROR), params.getTenantId()))
+            .orElseThrow(() -> new NotFoundException(String.format(
+              "Couldn't update failed jobExecutionSourceChunk status to ERROR, jobExecutionSourceChunk with id %s was not found", sourceChunkId))));
+        future.complete();
+      } else {
+        future.complete();
```
KaterynaSenchenko: complete future one time after the composite clause, and probably it would be nice to have some comment that this is done in order to continue data import process even if instance creation is failed

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -18,106 +15,40 @@
...
+                if (targetField.containsKey(AdditionalFieldsConfig.TAG_999)) {
+                  JsonObject instanceIdSubField = new JsonObject().put("i", instanceId);
+                  targetField.getJsonObject(AdditionalFieldsConfig.TAG_999).getJsonArray("subfields").add(instanceIdSubField);
+                  record.getParsedRecord().setContent(jsonContent.toString());
+                  return true;
+                }
+              }
+            }
+          } catch (Exception exception) {
+            LOGGER.error("Can not convert parsed record content to JsonObject. Cause:{}", exception.getMessage());
```
KaterynaSenchenko: please log the exception itself, not just its message

PR #106, opened 2019-05-17
Added Transfer encoding header

Igor-Gorchakov:


Igor-Gorchakov:
This approach doesn't work.

PR #111, opened 2019-05-23
MODSOURMAN-124 Added batch update of ParsedRecords after assigning Instance id to MA…

KaterynaSenchenko:
…RC records


mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/InstanceProcessingServiceImpl.java
```diff
@@ -62,8 +63,12 @@ public InstanceProcessingServiceImpl(@Autowired JobExecutionSourceChunkDao jobEx
             .map(sourceChunk -> jobExecutionSourceChunkDao.update(sourceChunk.withState(JobExecutionSourceChunk.State.ERROR), params.getTenantId()))
             .orElseThrow(() -> new NotFoundException(String.format(
               "Couldn't update failed jobExecutionSourceChunk status to ERROR, jobExecutionSourceChunk with id %s was not found", sourceChunkId))));
+      } else {
+        if (Record.RecordType.MARC.equals(records.get(0).getRecordType())) {
```
Anatolii-Starkov: Piece of advise. It's better to compare enums via `==` than equals method

PR #120, opened 2019-06-12
MODSOURMAN-131 - Remove stub data for landing page

RuslanLavrov:


NEWS.md
```diff
@@ -6,6 +6,7 @@
  * Applied parallel approach for mapping from records to instances
  * Use batch post to send Instances to the mod-inventory
  * MARC to Instance mapping was updated to the new one version
+ * Changed implementation for Job Execution human-readable id using DB sequence
```
KaterynaSenchenko: new iteration after release

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -119,7 +120,15 @@
   @Override
   public Future<String> save(JobExecution jobExecution, String tenantId) {
     Future<String> future = Future.future();
-    pgClientFactory.createInstance(tenantId).save(TABLE_NAME, jobExecution.getId(), jobExecution, future.completer());
+    String preparedQuery = String.format(GET_JOB_EXECUTION_HR_ID, PostgresClient.convertToPsqlStandard(tenantId));
+    pgClientFactory.createInstance(tenantId).select(preparedQuery, getHrIdAr -> {
+      if (getHrIdAr.succeeded()) {
+        jobExecution.setHrId(getHrIdAr.result().getResults().get(0).getLong(0).toString());
```
OleksiiKuzminov: plz add != null checks for getResults().get(0).getLong(0)

OleksiiKuzminov: and fail if null

PR #127, opened 2019-06-19
MODSOURMAN-139 JobExecution marked as Error if processing of at least one chunk failed

KaterynaSenchenko:


mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_processing_is_completed_and_has_error_function.sql
```diff
@@ -0,0 +1,24 @@
...
+SELECT count(_id) =
+	(SELECT count(_id)
+		FROM job_execution_source_chunks
+ 			WHERE (jsonb->>'jobExecutionId')::uuid = jobExecId)
+	as completed FROM
+	job_execution_source_chunks
+	WHERE (jsonb->>'jobExecutionId')::uuid = jobExecId
+		AND jsonb->>'state' IN ('COMPLETED', 'ERROR')
+UNION ALL
+(SELECT 0 < (SELECT count(_id)
```
Igor-Gorchakov: Try to SELECT count(id) > 0

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionSourceChunkDao.java
```diff
@@ -61,8 +62,9 @@
    *
    * @param jobExecutionId - UUID of related JobExecution
    * @param tenantId       - tenantId
-   * @return - return boolean value is all {@link JobExecutionSourceChunk} related to one jobExecution have status COMPLETED
+   * @return - returns a pair of boolean values, first indicates whether processing is completed,
+   * second - if there is any error
    */
-  Future<Boolean> isAllChunksProcessed(String jobExecutionId, String tenantId);
+  Future<Pair<Boolean, Boolean>> isAllChunksProcessed(String jobExecutionId, String tenantId);
```
Anatolii-Starkov: I suggest to return here StatusDto

mod-source-record-manager-server/src/main/java/org/folio/services/ChunkProcessingServiceImpl.java
```diff
@@ -44,18 +45,26 @@ public ChunkProcessingServiceImpl(@Autowired JobExecutionSourceChunkDao jobExecu
...
+      .compose(processingFinished -> {
+        // if processing completed
+        if (processingFinished.getLeft()) {
+          // if processing contains errors
+          if (processingFinished.getRight()) {
+            return checkAndUpdateJobExecutionStatusIfNecessary(jobExecutionId,
+              new StatusDto().withStatus(StatusDto.Status.ERROR).withErrorStatus(StatusDto.ErrorStatus.INSTANCE_CREATING_ERROR), params)
+              .compose(jobExecution -> jobExecutionService.updateJobExecution(jobExecution.withCompletedDate(new Date()), params))
+              .map(result -> true);
+          }
```
Anatolii-Starkov: If you get here a StatusDto you can avoid code duplication

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionSourceChunkDaoImpl.java
```diff
@@ -35,6 +35,7 @@
   public static final Logger LOGGER = LoggerFactory.getLogger(JobExecutionSourceChunkDaoImpl.class);
   private static final String TABLE_NAME = "job_execution_source_chunks";
   private static final String ID_FIELD = "'id'";
+  private static final String GET_PROCESSING_IS_COMPLETED_AND_HAS_ERROR_QUERY = "SELECT get_processing_is_completed_and_has_error('%s');";
```
Anatolii-Starkov: I would name it as GET_PROCESSING_STATUS. The "get" and "is" and "has" are a bit confusing me especially in one sentences

OleksiiKuzminov: In my opinion, it's an overengineering. There are no cases to do this solution badly. So I propose to close this one

PR #128, opened 2019-06-20
MODSOURMAN-130 - Update API documentation

RuslanLavrov:


README.md
```diff
@@ -189,11 +189,11 @@ curl -w '\n' -X POST -D -   \
 }
 ```
 
-### Post raw records to parsing
+### Post records to parsing
 
 To initiate records parsing should send POST request containing RawRecordsDto, wich contains raw records list ("records" field)
 to follow path: /change-manager/jobExecutions/{jobExecutionId}/records. 
-The list of records can contain records in different formats for example  "MARC_RAW", "MARC_JSON", "MARC_XML"
+The list of records can contains records in different formats for example:  "MARC_RAW", "MARC_JSON", "MARC_XML". \
```
Igor-Gorchakov: ... can contain ...

README.md
```diff
@@ -252,9 +279,10 @@ JobExecution entity will be have following changed state
 
 ### Finishing raw records parsing
 To indicate end of raw records transfering for parsing should send POST request containing last RawRecordsDto
-to follow path: /change-manager/jobExecutions/{jobExecutionId}/records. 
+to follow path: /change-manager/jobExecutions/{jobExecutionId}/records. \
 {jobExecutionId} - JobExecution id, which can be retrieved from response of JobExecution creation request.
-The last RawRecordsDto should contains empty raw records list ("records" field) and field "last" = true.
+The last RawRecordsDto should contains empty records list ("records" field), appropriate record format value
```
Igor-Gorchakov: Please describe an aim of the last request - A purpose of the last request is to confirm that there is not more records to upload for the current job. In this case request comes with "last" = true  sign and empty records inside.

PR #131, opened 2019-06-24
MODSOURMAN-132 Selective update of parsed records

Anatolii-Starkov:
-handle return values for save parsed records
-added request to response transformer in tests
-added tests


mod-source-record-manager-server/src/test/java/org/folio/dao/JobExecutionDaoImplUnitTest.java
```diff
@@ -20,12 +20,7 @@
 import java.util.UUID;
 
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
KaterynaSenchenko: please avoid wildcard imports

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -179,14 +177,16 @@ private void addAdditionalFieldsToMarcRecord(Record record) {
...
+
       RecordCollection recordCollection = new RecordCollection()
         .withRecords(parsedRecords)
         .withTotalRecords(parsedRecords.size());
+
       client.postSourceStorageBatchRecords(recordCollection, response -> {
-        if (response.statusCode() == HttpStatus.HTTP_CREATED.toInt()) {
-          future.complete(parsedRecords);
+        if (isStatus(response, HTTP_CREATED) || isPartialSuccess(response)) {
+          response.bodyHandler(it -> future.complete(it.toJsonObject().mapTo(RecordCollection.class).getRecords()));
```
OleksiiKuzminov: "it.toJsonObject().mapTo(RecordCollection.class)"
if there will be na error future will be never ended

OleksiiKuzminov:
I don't find logic where you update the record's only for created instances

mod-source-record-manager-server/src/test/java/org/folio/dao/JobExecutionDaoImplUnitTest.java
Anatolii-Starkov: done

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
Anatolii-Starkov: done

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -11,15 +14,7 @@
...
-import org.folio.rest.jaxrs.model.ErrorRecord;
-import org.folio.rest.jaxrs.model.JobExecution;
-import org.folio.rest.jaxrs.model.JobExecutionSourceChunk;
-import org.folio.rest.jaxrs.model.ParsedRecord;
-import org.folio.rest.jaxrs.model.RawRecord;
-import org.folio.rest.jaxrs.model.RawRecordsDto;
-import org.folio.rest.jaxrs.model.Record;
-import org.folio.rest.jaxrs.model.RecordCollection;
-import org.folio.rest.jaxrs.model.StatusDto;
+import org.folio.rest.jaxrs.model.*;
```
OleksiiKuzminov: imports

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -198,6 +202,29 @@ private void addAdditionalFieldsToMarcRecord(Record record) {
     return future;
   }
 
+  private Handler<Buffer> retrieveResponseHandler(Future<List<Record>> future) {
```
OleksiiKuzminov: Are we need to keep these methods here?

OleksiiKuzminov: looks like utility methods

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/InstanceProcessingServiceImpl.java
```diff
@@ -19,18 +12,19 @@
 import org.folio.dataimport.util.OkapiConnectionParams;
 import org.folio.dataimport.util.RestUtil;
 import org.folio.rest.client.SourceStorageBatchClient;
-import org.folio.rest.jaxrs.model.Instance;
-import org.folio.rest.jaxrs.model.Instances;
-import org.folio.rest.jaxrs.model.JobExecutionSourceChunk;
-import org.folio.rest.jaxrs.model.ParsedRecord;
-import org.folio.rest.jaxrs.model.ParsedRecordCollection;
-import org.folio.rest.jaxrs.model.Record;
+import org.folio.rest.jaxrs.model.*;
```
OleksiiKuzminov: import

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/InstanceProcessingServiceImpl.java
```diff
@@ -55,14 +49,22 @@ public InstanceProcessingServiceImpl(@Autowired JobExecutionSourceChunkDao jobEx
...
       }
       // Complete future in order to continue the import process regardless of the result of creating Instances
       future.complete();
     });
     return future;
   }
 
+  private List<Pair<Record, Instance>> calculateRecordsToUpdate(List<Pair<Record, Instance>> recordToInstanceList, List<Instance> result) {
+    return recordToInstanceList.stream()
+      .filter(it -> result.contains(it.getValue()))
```
OleksiiKuzminov: contains works with equals. Is there working fine? We can use only id field to understand that instance is the same

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/AbstractRestTest.java
```diff
@@ -215,11 +217,65 @@ protected InitJobExecutionsRsDto constructAndPostInitJobExecutionRqDto(int files
...
       .spec(spec)
       .body(JsonObject.mapFrom(statusDto).toString())
       .when()
       .put(JOB_EXECUTION_PATH + jobExecution.getId() + "/status");
   }
+
+  /**
+   * Maps a request body to a response body.
+   */
+  public static class RequestToResponseTransformer extends ResponseTransformer {
```
OleksiiKuzminov: Cool stuff! looks like a good idea to use!

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
OleksiiKuzminov: not fixed

PR #133, opened 2019-06-25
MODSOURCE-58 - Applied auto-generation in record-manager

RuslanLavrov:


OleksiiKuzminov:
is there an actual version of mod-source-record-storage client?

RuslanLavrov:
> is there an actual version of mod-source-record-storage client?
yes, 2.5.0-SNAPSHOT

PR #138, opened 2019-07-02
MODSOURMAN-142 Check for processing errors is separated into different method

KaterynaSenchenko:


mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionSourceChunkDaoImpl.java
```diff
@@ -35,7 +34,8 @@
   public static final Logger LOGGER = LoggerFactory.getLogger(JobExecutionSourceChunkDaoImpl.class);
   private static final String TABLE_NAME = "job_execution_source_chunks";
   private static final String ID_FIELD = "'id'";
-  private static final String GET_PROCESSING_STATE_QUERY = "SELECT get_processing_state('%s');";
+  private static final String GET_PROCESSING_STATE_QUERY = "SELECT is_processing_completed('%s');";
+  private static final String HAS_ERRORS_QUERY = "SELECT has_errors('%s');";
```
OleksiiKuzminov: naming of function and var should be more clear. Who has errors?

mod-source-record-manager-server/src/main/java/org/folio/services/ChunkProcessingServiceImpl.java
```diff
@@ -122,25 +121,29 @@ public ChunkProcessingServiceImpl(@Autowired JobExecutionSourceChunkDao jobExecu
...
+                future.complete(new StatusDto().withStatus(StatusDto.Status.ERROR).withErrorStatus(StatusDto.ErrorStatus.INSTANCE_CREATING_ERROR));
+              } else {
+                // status should be JobExecution.Status.PARSING_FINISHED but for first version we finish import in this place
+                future.complete(new StatusDto().withStatus(StatusDto.Status.COMMITTED));
+              }
+              return future;
+            });
         }
-        return new StatusDto().withStatus(StatusDto.Status.PARSING_IN_PROGRESS);
+        return Future.succeededFuture(new StatusDto().withStatus(StatusDto.Status.PARSING_IN_PROGRESS));
```
Anatolii-Starkov: Could you refactor it
sort of 
```
/**
   * Checks actual status of JobExecution
   *
   * @param jobExecutionId - JobExecution id
   * @return future with actual JobExecution status
   */
  private Future<StatusDto> checkJobExecutionState(String jobExecutionId, String tenantId) {
    return jobExecutionSourceChunkDao.get("jobExecutionId=" + jobExecutionId + " AND last=true", 0, 1, tenantId)
      .compose(isAllChunkProcessed(jobExecutionId, tenantId))
      .compose(hasSomeChunkError(jobExecutionId, tenantId));
  }

  private Function<Boolean, Future<StatusDto>> hasSomeChunkError(String jobExecutionId, String tenantId) {
    return completed -> {
      if (completed) {
        return jobExecutionSourceChunkDao.containsErrorChunks(jobExecutionId, tenantId)
          .map(hasErrors -> succeededFuture(hasErrors ? statusError() : statusCommitted()))
          .result();
      } else {
        return succeededFuture(statusParsingInProgress());
      }
    };
  }

  private Function<List<JobExecutionSourceChunk>, Future<Boolean>> isAllChunkProcessed(String jobExecutionId, String tenantId) {
    return chunks -> isNotEmpty(chunks) ? jobExecutionSourceChunkDao.isAllChunksProcessed(jobExecutionId, tenantId) : succeededFuture(false);
  }

  private StatusDto statusParsingInProgress() {
    return new StatusDto().withStatus(StatusDto.Status.PARSING_IN_PROGRESS);
  }

  private StatusDto statusCommitted() {
    return new StatusDto().withStatus(StatusDto.Status.COMMITTED);
  }

  private StatusDto statusError() {
    return new StatusDto()
      .withStatus(StatusDto.Status.ERROR)
      .withErrorStatus(StatusDto.ErrorStatus.INSTANCE_CREATING_ERROR);
  }
```

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionSourceChunkDao.java
```diff
@@ -57,14 +57,23 @@
   Future<Boolean> delete(String id, String tenantId);
 
   /**
-   * Verifies is all related chunks (to given job execution id) are processed
+   * Checks whether all related chunks (to given job execution id) are processed
    * Each chunk is considered being processed if it has ERROR or COMPLETED state.
```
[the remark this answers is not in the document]
KaterynaSenchenko: Well, chunk processing can finish either successfully (COMPLETED status) or with error (ERROR status). We consider the whole process finish when all the chunks are processed.

mod-source-record-manager-server/src/main/java/org/folio/services/ChunkProcessingServiceImpl.java
KaterynaSenchenko: Thanks for the suggestions, I refactored the method in a slightly different manner

Anatolii-Starkov: 1. Ternary operator must be in one line and width of it should be 120 characters. In other case it 's difficult to read.
2. Comments should be in javadoc or meaning can be express by name of a method
3. More then one return is bad coding style 

I suggest you to use private methods to avoid a mess in production code especially when you use a bunch of compose methods

KaterynaSenchenko: I beg to differ,
1. in this case separating a block in "helper" private method that returns a function does not help in readability
2. in my opinion, in-line comments are acceptable in general. In this case we have to explain the choice of the status as it will change in future.
3. agree, but sometimes it makes for better readability.

Asynchronous code is hard to read by definition, we moved from "callback hell" to composing style to improve readability, but further decomposition on "helper" private methods, in my opinion, does not help, especially when in concerns wrapping builder methods. 

Assigning this PR to @OleksiiKuzminov to resolve the argument.

KaterynaSenchenko: 3 people approved this PR, so I'm merging it. If you still dissatisfied @Anatolii-Starkov, please create a separate JIRA issue for further refactoring. 
CC: @OleksiiKuzminov

PR #146, opened 2019-07-18
MODDATAIMP-149 - data-import display error on FOLIO stable hosted environment

RuslanLavrov:


mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -224,6 +227,7 @@ private JobExecution buildNewJobExecution(boolean isParent, boolean isSingle, St
       .withHrId(String.valueOf(random.nextInt(99999)))
       .withParentJobId(parentJobExecutionId)
       .withSourcePath(fileName)
+      .withRunBy(new RunBy().withFirstName("DIKU").withLastName("ADMINISTRATOR"))
```
OleksiiKuzminov: plz format

PR #160, opened 2019-08-27
MODDATAIMP-136 - Updates to default MARC-to-Instance mapping

Igor-Gorchakov:
1. Identifiers: Valid / Invalid ISBN
2. Identifiers: ISSN, Invalid ISSN, Linking ISSN
3. Identifiers: System control number vs. OCLC number vs. Cancelled system control numbers
4. Identifiers: GPO and Cancelled GPO number
5. Languages
6. Classification
7. Index title
8. Resource type (no mapping, only added a flag into Processor to skip repeatable fields)

Rework for rules Processor:
* Normalization Functions accept parameters as JsonObject (it was String before);
* Normalization Functions accept RuleProcessingContext with MARC DataField and subfield Data;
* Added new parameter _requiredSubfield_ (see rules.json). If incoming MARC record has some **field** with sub-fields, but does not have sub-field that is in _requiredSubfield_ parameter, then Processor skips mapping for the given **field**;
* Added new parameter  _ignoreSubsequentFields_. If some rule
 has `"ignoreSubsequentFields" : true` for some **Field**, then Processor skips mapping for other fields with the same tag


mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -1,19 +1,31 @@
...
+
+import static org.drools.core.util.StringUtils.EMPTY;
 
 /**
  * Run a splitter on a string or run a function.
  */
 public class NormalizationFunctions {
+  private static final Logger LOGGER = LoggerFactory.getLogger(NormalizationFunctions.class);
 
   private static final String CHAR_SELECT = "char_select";
```
OleksiiKuzminov: Maybe struct this constants and create enums?

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -23,91 +35,120 @@ private NormalizationFunctions() {
...
-    }
-    else if(TRIM.equals(funcName)){
-      return val.trim();
+  public static String runFunction(String functionName, RuleExecutionContext ruleExecutionContext, JsonObject parameters) {
+    String subFieldData = ruleExecutionContext.getData();
+    try {
+      if (subFieldData == null) {
+        return EMPTY;
+      }
+      if (CHAR_SELECT.equals(functionName)) {
```
OleksiiKuzminov: switch case ?

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -23,91 +35,120 @@ private NormalizationFunctions() {
...
+      return subFieldData;
     }
-    else if(TRIM_PERIOD.equals(funcName)){
-      return trimPeriod(val);
+    return EMPTY;
+  }
+
+  private static String setValueIfSubfieldHasPrefix(String subFieldData, JsonObject parameters) {
+    List<String> prefixes = parameters.getJsonArray("prefixes").getList();
+    String valueIfConditionIsTrue = parameters.getString("valueIfConditionIsTrue");
```
OleksiiKuzminov: move string to variables

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/NormalizationFunctions.java
```diff
@@ -23,91 +35,120 @@ private NormalizationFunctions() {
...
-        else if(input.substring(input.length()-3).equals("...")){
-          return input;
-        }
-        else {
+  private static String modified(final String subFieldData, int pos) {
+    if (PUNCT_2_REMOVE.contains(String.valueOf(subFieldData.charAt(pos)))) {
+      return subFieldData.substring(0, subFieldData.length() - 1);
+    } else if (subFieldData.charAt(pos) == '.') {
+      try {
+        if (subFieldData.substring(subFieldData.length() - 4).equals("....")) {
```
OleksiiKuzminov: dots to a named variable. Hard to understand what it is

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/Processor.java
```diff
@@ -100,36 +101,50 @@ private void processDataFieldSection(Iterator<DataField> dfIter) throws IllegalA
...
       //per subfield in the marc field
       JsonObject subFieldMapping = mappingEntry.getJsonObject(i);
-      processSubFieldMapping(subFieldMapping, rememberComplexObj, dataField);
+      if (canProcessSubFieldMapping(subFieldMapping, dataField)) {
+        processSubFieldMapping(subFieldMapping, rememberComplexObj, dataField);
+      }
     }
   }
 
+  private boolean canProcessSubFieldMapping(JsonObject subFieldMapping, DataField dataField) {
```
OleksiiKuzminov: java doc

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/RuleExecutionContext.java
```diff
@@ -0,0 +1,25 @@
+package org.folio.services.mappers.processor;
+
+import org.marc4j.marc.DataField;
+
+public class RuleExecutionContext {
```
OleksiiKuzminov: java docs

PR #165, opened 2019-08-31
MODSOURMAN-168 - Integrate rule processor with Settings

Igor-Gorchakov:
Checked on local vagrant
image


mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/InstanceProcessingServiceImpl.java
```diff
@@ -66,21 +68,25 @@ public InstanceProcessingServiceImpl(@Autowired JobExecutionSourceChunkDao jobEx
...
-      }
-      updateSourceChunkState(sourceChunkId, sourceChunkState, params)
-        .compose(updatedChunk -> jobExecutionSourceChunkDao.update(updatedChunk.withCompletedDate(new Date()), params.getTenantId()))
-        // Complete future in order to continue the import process regardless of the result of creating Instances
-        .setHandler(updateAr -> future.complete());
-    });
+    MappingParametersBuilder.build(params)
+      .compose(mappingParameters -> {
+        Map<Instance, Record> instanceRecordMap = mapRecords(records, mappingParameters, params);
+        List<Instance> instances = new ArrayList<>(instanceRecordMap.keySet());
```
OleksiiKuzminov: plz move to the method

PR #174, opened 2019-09-10
Modsourman 166

OleksiiKuzminov:
### Steps

- Load Resource type from inventory settings
- Create a new function in Processor
- Use function in rule.json

Mapping rules Resource type :
Comes from the 336 field of the MARC record
Read the values in $a $b and $2 respectively and map to Resource type Term, Code, and Source
Note: MARC 336 field is repeatable. Instance Resource type is not repeatable. If there are multiple 336 fields in the MARC record, only map the first one, and ignore the subsequent ones.

### DIFF
--- a/mod-source-record-manager-server/src/main/resources/rules.json
+++ b/mod-source-record-manager-server/src/main/resources/rules.json
@@ -630,7 +630,8 @@
           "target": "identifiers.identifierTypeId",
           "description": "Type for System Control Number",
           "subfield": [
-            "a", "z"
+            "a",
+            "z"
           ],
           "rules": [
             {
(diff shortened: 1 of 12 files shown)


mod-source-record-manager-server/src/main/resources/rules.json
```diff
@@ -1565,6 +1579,26 @@
       ]
     }
   ],
+  "336": [
+    {
+      "target": "instanceTypeId",
+      "description": "Instance type ID",
+      "ignoreSubsequentFields": true,
+      "applyRulesOnConcatenatedData": true,
```
Igor-Gorchakov: Remove "applyRulesOnConcatenatedData" flag

Igor-Gorchakov:
Please cover new function with unit test, see `NormalizationFunctionTest`

PR #179, opened 2019-09-16
Modsourman 187 contributors type

Igor-Gorchakov:
**Contributor Name**: in 100, 110, 111, 700, 710, or 711 field
**Contributor Type**: Coded version is noted in $4
**Contributor Primary**: If contributor comes from 100, 110, or 111 field, mark it as Primary (true)
If contributor comes from 700, 710, or 711 field, do not mark it as Primary (false)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
@@ -4,6 +4,7 @@ import io.vertx.core.json.JsonObject;
 import org.apache.commons.lang.StringUtils;
 import org.folio.rest.jaxrs.model.InstanceType;
 import org.folio.rest.jaxrs.model.ClassificationType;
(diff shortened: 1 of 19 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
```diff
@@ -161,6 +162,21 @@ public String apply(RuleExecutionContext context) {
...
 
+  SET_CONTRIBUTOR_TYPE_ID() {
+    @Override
+    public String apply(RuleExecutionContext context) {
+      List<ContributorType> types = context.getMappingParameters().getContributorTypes();
+      if (types == null) {
+        return StringUtils.EMPTY;
+      }
+      return types.stream()
+        .filter(instanceType -> instanceType.getCode().equals(context.getSubFieldValue()))
```
OleksiiKuzminov: contributerType

Igor-Gorchakov: Will fix in next PR

PR #183, opened 2019-09-19
MODSOURMAN-194 Added mapping for GPO and Cancelled GPO number identifier types

KaterynaSenchenko:
Identifier: GPO and Cancelled GPO number (rows 14, 15)
Review the subfields carefully to determine GPO Identifier Type
074 $a = [Valid] GPO item number
074 $z = Cancelled GPO item number
Each of the above subfields should be in a separate Instance GPO number field, even if they are in the same MARC 074 field

Examples
074 \\$a1002-A
074 \\$a1002-B (MF)
Become
GPO item number 1002-A
GPO item number 1002-B (MF)

074 \\$a1022-A$z1012-A
Become
GPO item number 1022-A
Cancelled GPO item number 1012-A


Igor-Gorchakov:
NEWS ??

PR #184, opened 2019-09-19
MODSOURMAN-170: setup isbn type id from settings

OleksiiKuzminov:
ISBN
Comes from 020 $a $q $c all put together in whatever order they appear in the 020; $q may be repeated; remove the subfield mark and letter; add a space between subfields

Invalid ISBN 
Comes from 020 $z $q $c all put together in whatever order they appear in the 020; $q may be repeated; remove the subfield mark and letter; add a space between subfields

### DIFF
--- a/mod-source-record-manager-server/src/main/resources/rules.json
+++ b/mod-source-record-manager-server/src/main/resources/rules.json
@@ -359,8 +359,14 @@
           ],
           "rules": [
             {
-              "conditions": [],
-              "value": "8261054f-be78-422d-bd51-4ed9f33c3422"
+              "conditions": [
+                {
+                  "type": "set_identifier_type_id_by_name",
+                  "parameter": {
+                    "name": "ISBN"
+                  }
+                }
+              ]
             }
           ]
         },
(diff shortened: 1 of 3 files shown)


mod-source-record-manager-server/src/main/resources/rules.json
```diff
@@ -359,8 +359,14 @@
           ],
```
Igor-Gorchakov:
```suggestion
          "q",
          "c"
          ],
```

Igor-Gorchakov: subfields :  
"a", "c", "q"   ?

mod-source-record-manager-server/src/main/resources/rules.json
```diff
@@ -359,8 +359,14 @@
...
           "rules": [
```
Igor-Gorchakov: "applyRulesOnConcatenatedData": true

mod-source-record-manager-server/src/main/resources/rules.json
OleksiiKuzminov: yes

mod-source-record-manager-server/src/main/resources/rules.json
```diff
@@ -398,8 +404,14 @@
           ],
```
Igor-Gorchakov: subfields: "z",
            "q",
            "c"   ?

Igor-Gorchakov: "applyRulesOnConcatenatedData": true,

PR #196, opened 2019-10-06
Modsourman 181 CLI

Igor-Gorchakov:


mod-source-record-manager-server/src/main/java/org/folio/services/MappingRuleServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+    return future;
+  }
+
+  /**
+   * Returns true if given String is valid JSON
+   *
+   * @param json given string
+   * @return true if given String is valid JSON, false if non-valid
+   */
+  private boolean isValidJson(String json) {
```
Igor-Gorchakov: @RuslanLavrov , @KaterynaSenchenko , @OleksiiKuzminov     please help me to find more useful way to check String if it's in Json format or not

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -35,30 +40,42 @@
...
   }
 
-  private Future<UpdateResult> setSequencesPermissionForDbUser(Map<String, String> headers, Context context) {
+  private Future<UpdateResult> setSequencesPermissionForDbUser(Context context, String tenantId) {
     Future<UpdateResult> future = Future.future();
-    String dbSchemaName = new StringBuilder(TenantTool.calculateTenantId(headers.get("x-okapi-tenant")))
-      .append('_').append(PostgresClient.getModuleName()).toString();
+    String dbSchemaName = new StringBuilder()
+      .append(tenantId)
+      .append('_')
```
KaterynaSenchenko: ('_')

KaterynaSenchenko: (*_*)

PR #215, opened 2019-11-08
Fix grammar for documentation for rules processor

Igor-Gorchakov:


RuleProcessorApi.md
```diff
@@ -53,9 +53,9 @@ Instance:
 }
 ```
 
-[Here](https://github.com/folio-org/mod-source-record-manager/blob/master/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java) all the formatting functions defined. Most useful are: `trim, capitalize, remove_ending_punc`.
+[Here](https://github.com/folio-org/mod-source-record-manager/blob/master/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java) are all of the formatting functions defined. The most useful ones are. The most useful ones are: `trim, capitalize, remove_ending_punc`.
```
Igor-Gorchakov: TODO // remove doubles

RuleProcessorApi.md
```diff
@@ -53,9 +53,9 @@ Instance:
...
 
- In most cases there are sub-fields present in field, that is important for mapping. Example for "250" field with `a, b, 6` sub-fields comes below: 
+In most cases there are subfields present within MARC fields. This is important for mapping. Here is an example of a MARC “250” field that contains a “subfield a”, a “subfield b”, and a “subfield 6.”
```
Igor-Gorchakov: TODO: add path to Instance schema

RuleProcessorApi.md
```diff
@@ -78,15 +78,15 @@ Rule:
...
 {
   "edition": "Fifth ed. Editor in chief Lord Mackay of Clashfern."
 }
 ```
 #### Mapping for complex fields
-What if the target Instance field is not simple String, but List of complex objects with several fields in ? This happends usually if record field is a **Data field**. We can write rule to map record as below: 
+What if there is a separate target in an Instance record for each subfield within a MARC field? If that is the case, then we can write the mapping rules as shown below:
  ```json
 MARC Record: "264":{"subfields":[{"a":"Chicago, Illinois :"}, {"b":"The HistoryMakers,"}, {"c":"[2016]"}], "ind1":" ", "ind2":"1"}
 ```
```
Igor-Gorchakov: TODO put a link to publication schema

RuleProcessorApi.md
```diff
@@ -154,7 +153,7 @@ Instance:
...
 
 #### Multiple objects from one field
```
Igor-Gorchakov: TODO remove 'f', 'g', 'h' subfields. Find real examples

RuleProcessorApi.md
```diff
@@ -419,13 +418,30 @@ There are 3 REST methods to work with rules.
 |**PUT**| /mapping-rules | application/json | Update rules for given tenant |
 |**PUT**| /mapping-rules/restore | application/json | Restore rules to default |
 
-To get rules you can send this request using GET method
+Before working with API make sure you have an HTTP token that is required for sending requests. If you have already logged in the system using UI, just copy token from `Apps/Settings/Developer/Set token/Authentication token` field.
+Also you can log in the system using CLI tools, response returns `x-okapi-token` header with HTTP token.
```
Igor-Gorchakov: TODO CLI tools like CURL or PostMan

PR #217, opened 2019-11-13
MODSOURMAN-221 - Implement journal service and apply it data-import process

RuslanLavrov:


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -52,11 +58,13 @@
 
   private JobExecutionSourceChunkDao jobExecutionSourceChunkDao;
   private JobExecutionService jobExecutionService;
+  private JournalService journalService;
 
   public ChangeEngineServiceImpl(@Autowired JobExecutionSourceChunkDao jobExecutionSourceChunkDao,
                                  @Autowired JobExecutionService jobExecutionService) {
     this.jobExecutionSourceChunkDao = jobExecutionSourceChunkDao;
     this.jobExecutionService = jobExecutionService;
+    this.journalService = JournalService.createProxy(Vertx.currentContext().owner());
```
OleksiiKuzminov: @autowire Vertx?

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -44,6 +46,7 @@
 import static java.lang.String.format;
 import static org.folio.HttpStatus.HTTP_CREATED;
 import static org.folio.rest.RestVerticle.MODULE_SPECIFIC_ARGS;
+import static org.folio.rest.jaxrs.model.JournalRecord.ActionType.*;
```
OleksiiKuzminov: import

PR #218, opened 2019-11-18
Preferences existing 999, if present, record id

jeremythuff:
Resolves [MODSOURMAN-233](https://issues.folio.org/projects/MODSOURMAN/issues/MODSOURMAN-233):

The intent of this PR is to allow an incoming raw marc to specify a 999 field, subfield i. Once specified this fields value will be used for the record identifier, instead of a random UUID.


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -151,6 +153,30 @@ public ChangeEngineServiceImpl(@Autowired JobExecutionSourceChunkDao jobExecutio
...
+    String id = "";
+    Optional<JsonObject> nineNineNineField = findInArrayByKey(fields, "999");
+    if(nineNineNineField.isPresent()) {
+      JsonArray subFields = nineNineNineField.get().getJsonArray("subfields", new JsonArray());
+      Optional<JsonObject> subFieldI = findInArrayByKey(subFields, "i");
+      if(subFieldI.isPresent()) {
+        id = subFieldI.get().getString("i");
+      }
+    }
+    return StringUtils.isEmpty(id) ? UUID.randomUUID().toString() : id;
```
OleksiiKuzminov: What should be in case when try to load two records with the same 999 field id or we try to load it twice?

jeremythuff: Am I correct in assuming that the current behavior, when loading an identical record multiple  times, is for duplicate records to be created with different UUIDs? If so, is this behavior explicitly desirable? 

It is the case that the introduction of this PR would likely result in the failure to create a duplicate record. That, however, does seem consistent with the behavior of most of the storage modules.

OleksiiKuzminov: You right, but we should store different versions of the same record. Snapshot id and generation fields will help us with it. So the approach to store this id in separate field is more preferable.

jeremythuff: Are you suggesting that rather than $i, this PR should be amended to use a a different subfield (e.g. $j) to express the UUID to be used as the Record ID?

jeremythuff: Or, are you suggesting that the [rawRecordDto](https://github.com/folio-org/data-import-raml-storage/blob/f4eceb9d60c08179257fd67de9e558cb67658858/schemas/dto/rawRecordDto.json) be emended to include an instanceID property?

If that is the case, I feel that I like the possibility of that solution much better.

OleksiiKuzminov: Yes, it is a good idea. But im not sure that instanceId is good name for this field. Can we store in the same way a holding or marcbib ids? We already have these ids in exteranalIdHolder field
https://github.com/folio-org/data-import-raml-storage/blob/7fcef21eb07b65a334447e8e2dcb5c77c825f00e/schemas/mod-source-record-storage/recordModel.json

PR #228, opened 2020-01-06
MODSOURMAN-246 Instance creation fails with default instancetypeid

Igor-Gorchakov:


mod-source-record-manager-server/src/main/resources/rules/rules.json
```diff
@@ -10,248 +10,12 @@
...
-              "type": "char_select",
-              "parameter": {
-                "from": 6,
-                "to": 7
-              },
-              "value": "a",
-              "LDR": true
-            }
-          ],
-          "value": "6312d172-f0cf-40f6-b27d-9fa8feaf332f"
```
Igor-Gorchakov: These ids are wrong, they were just overridden by rule 336 while mapping. Now we use id of the Unspecified type and override it by 336 rule

PR #232, opened 2020-01-24
MODSOURMAN-255: Logs show that failed to load files contain 100 records

VRohach:
JIRA-ticket: https://issues.folio.org/browse/MODSOURMAN-255

1. If JobExecution failed - the value of the total records will be 0.
2. The test was updated.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
@@ -155,11 +155,7 @@ public class JobExecutionServiceImpl implements JobExecutionService {
           } else {
             jobExecution.setStatus(JobExecution.Status.fromValue(status.getStatus().name()));
             jobExecution.setUiStatus(JobExecution.UiStatus.fromValue(Status.valueOf(status.getStatus().name()).getUiStatus()));
-            if (status.getStatus().equals(StatusDto.Status.ERROR)) {
-              jobExecution.withErrorStatus(JobExecution.ErrorStatus.fromValue(status.getErrorStatus().name()))
-                .withProgress(jobExecution.getProgress().withTotal(0))
-                .withCompletedDate(new Date());
-            }
+            updateJobExecutionIfErrorExist(status, jobExecution);
             future.complete(jobExecution);
           }
         } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -157,6 +157,7 @@
             jobExecution.setUiStatus(JobExecution.UiStatus.fromValue(Status.valueOf(status.getStatus().name()).getUiStatus()));
             if (status.getStatus().equals(StatusDto.Status.ERROR)) {
               jobExecution.withErrorStatus(JobExecution.ErrorStatus.fromValue(status.getErrorStatus().name()))
+                .withProgress(jobExecution.getProgress().withTotal(0))
```
KaterynaSenchenko: Please update progress in place where errorStatus is set for JobExecution (check if it is FILE_PROCESSING_ERROR).

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -443,4 +439,19 @@ private Future saveSnapshots(List<Snapshot> snapshots, OkapiConnectionParams par
...
+   * Updates jobExecution object, if Error exists.
+   * @param status - DTO which contains new status
+   * @param jobExecution - specific JobExecution
+   */
+  private void updateJobExecutionIfErrorExist(StatusDto status, JobExecution jobExecution) {
+    if (status.getStatus().equals(StatusDto.Status.ERROR)) {
+      jobExecution.withErrorStatus(JobExecution.ErrorStatus.fromValue(status.getErrorStatus().name()))
+        .withCompletedDate(new Date());
+      if(jobExecution.getErrorStatus().equals(JobExecution.ErrorStatus.FILE_PROCESSING_ERROR)){
+        jobExecution.withProgress(jobExecution.getProgress().withTotal(0));
```
KaterynaSenchenko: pls use setter here as well

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -443,4 +439,19 @@ private Future saveSnapshots(List<Snapshot> snapshots, OkapiConnectionParams par
...
     return future;
   }
+
+  /**
+   * Updates jobExecution object, if Error exists.
+   * @param status - DTO which contains new status
+   * @param jobExecution - specific JobExecution
+   */
+  private void updateJobExecutionIfErrorExist(StatusDto status, JobExecution jobExecution) {
+    if (status.getStatus().equals(StatusDto.Status.ERROR)) {
```
KaterynaSenchenko: just a note, we can use '==' with enums

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -443,4 +439,19 @@ private Future saveSnapshots(List<Snapshot> snapshots, OkapiConnectionParams par
...
+      jobExecution.withErrorStatus(JobExecution.ErrorStatus.fromValue(status.getErrorStatus().name()))
```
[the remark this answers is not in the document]
VRohach: Sure, I will do it. It was just copied from the previous place where this logic situated before extracting to this method

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
VRohach: Done

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
VRohach: Done

VRohach:
> Could you also add a test for this check? Thanks

Done. Could you please review it ?

PR #233, opened 2020-01-27
MODSOURMAN-206 HRID handling in SRS for MARC bib records: Create

OleksiiKuzminov:


mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/InstanceProcessingServiceImpl.java
```diff
@@ -69,20 +69,23 @@
...
     this.mappingParametersProvider = mappingParametersProvider;
     this.mappingRuleService = mappingRuleService;
     this.journalService = JournalService.createProxy(vertx);
+    this.hrIdFieldService = hrIdFieldService;
   }
 
   @Override
-  public Future<Void> process(List<Record> records, String sourceChunkId, OkapiConnectionParams okapiParams) {
-    Future future = Future.future();
+  public Promise<Void> process(List<Record> records, String sourceChunkId, OkapiConnectionParams okapiParams) {
```
RuslanLavrov: please change the method signature to return Future object instead of a Promise object

PR #235, opened 2020-02-07
MODSOURMAN-242 - Implement event sending on saving records in record-storage

RuslanLavrov:


mod-source-record-manager-server/src/main/java/org/folio/services/EventDrivenChunkProcessingServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+   * @return future with boolean
+   */
+  private Future<Boolean> sendEventsWithCreatedRecords(List<Record> createdRecords, String jobExecutionId, OkapiConnectionParams params) {
+    return jobExecutionService.getJobExecutionById(jobExecutionId, params.getTenantId())
+      .compose(jobOptional -> jobOptional
+        .map(jobExecution -> getJobProfileSnapshotWrapper(jobExecution, params))
+        .orElse(Future.failedFuture(new NotFoundException(String.format("Couldn't find JobExecution with id %s", jobExecutionId)))))
+      .compose(profileSnapshotWrapper -> {
+        List<Future> futures = createdRecords.stream()
+          .filter(record -> record.getParsedRecord() != null && record.getParsedRecord().getContent() != null)
```
OleksiiKuzminov: plz dont skip these records, log them

mod-source-record-manager-server/src/main/java/org/folio/services/EventDrivenChunkProcessingServiceImpl.java
```diff
@@ -0,0 +1,170 @@
...
+   *
+   * @param createdRecords records to send
+   * @param jobExecutionId job execution id
+   * @param params         connection parameters
+   * @return future with boolean
+   */
+  private Future<Boolean> sendEventsWithCreatedRecords(List<Record> createdRecords, String jobExecutionId, OkapiConnectionParams params) {
+    return jobExecutionService.getJobExecutionById(jobExecutionId, params.getTenantId())
+      .compose(jobOptional -> jobOptional
+        .map(jobExecution -> getJobProfileSnapshotWrapper(jobExecution, params))
```
KaterynaSenchenko: Let's get JobProfileWrapper once and store it in JobExecution to eliminate unnecessary calls to mod-data-import-converter-storage on every chunk.

PR #240, opened 2020-02-20
MODSOURCE-108 - Define rules for instance identifier types without hardcoded types UUIDs

RuslanLavrov:
…UIDs


OleksiiKuzminov:
plz update documentation

PR #242, opened 2020-02-24
MODSOURMAN-203: Map Mode of issuance

VRohach:
1. New mechanism for mapping "mode of issuance" was created.
2. Mode of issuance entity was created.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
@@ -12,8 +12,10 @@ import org.folio.rest.jaxrs.model.IdentifierType;
 import org.folio.rest.jaxrs.model.InstanceFormat;
 import org.folio.rest.jaxrs.model.InstanceNoteType;
 import org.folio.rest.jaxrs.model.InstanceType;
+import org.folio.rest.jaxrs.model.IssuanceMode;
 import org.folio.services.mappers.processor.RuleExecutionContext;
 import org.folio.services.mappers.processor.functions.enums.ElectronicAccessRelationshipEnum;
+import org.folio.services.mappers.processor.functions.enums.IssuanceModeEnum;
 import org.folio.services.mappers.processor.publisher.PublisherRole;
 import org.marc4j.marc.DataField;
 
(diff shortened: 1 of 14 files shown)


VRohach:
Tests for this implementation currently in progress

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
```diff
@@ -351,6 +353,52 @@ public String apply(RuleExecutionContext context) {
...
         .orElse(STUB_FIELD_TYPE_ID);
     }
+  },
+
+  SET_ISSUANCE_MODE_ID() {
+
+    @Override
+    public String apply(RuleExecutionContext context) {
+      String subFieldValue = context.getSubFieldValue();
+      char seventhChar = subFieldValue.charAt(6);
```
OleksiiKuzminov: comment

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
```diff
@@ -351,6 +353,52 @@ public String apply(RuleExecutionContext context) {
...
+    private String findIssuanceModeId(List<IssuanceMode> issuanceModes, IssuanceModeEnum issuanceModeType,
+                                      String defaultId) {
+      return issuanceModes.stream()
+        .filter(issuanceMode -> issuanceMode.getName().equals(issuanceModeType.getValue()))
+        .findFirst()
+        .map(IssuanceMode::getId)
+        .orElse(defaultId);
+    }
+
+    private String matchIssuanceModeIdViaLeaderSymbol(char seventhChar, List<IssuanceMode> issuanceModes, String defaultId) {
```
OleksiiKuzminov: move to enum

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
```diff
@@ -351,6 +353,52 @@ public String apply(RuleExecutionContext context) {
...
+        return StringUtils.EMPTY;
+      }
+      String defaultIssuanceModeId = findIssuanceModeId(issuanceModes, IssuanceModeEnum.UNSPECIFIED, StringUtils.EMPTY);
+      return matchIssuanceModeIdViaLeaderSymbol(seventhChar, issuanceModes, defaultIssuanceModeId);
+    }
+
+    private String findIssuanceModeId(List<IssuanceMode> issuanceModes, IssuanceModeEnum issuanceModeType,
+                                      String defaultId) {
+      return issuanceModes.stream()
+        .filter(issuanceMode -> issuanceMode.getName().equals(issuanceModeType.getValue()))
```
OleksiiKuzminov: equalsIgoneCase

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
```diff
@@ -351,6 +353,52 @@ public String apply(RuleExecutionContext context) {
...
+
+  SET_ISSUANCE_MODE_ID() {
+
+    @Override
+    public String apply(RuleExecutionContext context) {
+      String subFieldValue = context.getSubFieldValue();
+      char seventhChar = subFieldValue.charAt(6);
+      List<IssuanceMode> issuanceModes = context.getMappingParameters().getIssuanceModes();
+      if (issuanceModes == null || issuanceModes.isEmpty()) {
+        return StringUtils.EMPTY;
```
OleksiiKuzminov: log debug cases with StringUtils.EMPTY

OleksiiKuzminov:
unit tests

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
```diff
@@ -351,6 +353,52 @@ public String apply(RuleExecutionContext context) {
...
+  },
+
+  SET_ISSUANCE_MODE_ID() {
+
+    @Override
+    public String apply(RuleExecutionContext context) {
+      String subFieldValue = context.getSubFieldValue();
+      char seventhChar = subFieldValue.charAt(6);
+      List<IssuanceMode> issuanceModes = context.getMappingParameters().getIssuanceModes();
+      if (issuanceModes == null || issuanceModes.isEmpty()) {
```
KaterynaSenchenko:
```suggestion
      if (CollectionUtils.isEmpty(issuanceModes)) {
```

VRohach: I implement this check like in other Normalization-functions for common approach. But I think that `CollectionUtils.isEmpty()` -  is the best way to do it. So, I will change it in other normalization-function for saving a common approach. Thx @KaterynaSenchenko !

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
VRohach: Yes, sure. I added debug logging for each case where we will have an empty string as a result of normalization function.

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
VRohach: Done

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
VRohach: Done. @OleksiiKuzminov could you please review my approach?

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
VRohach: Added.

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/functions/NormalizationFunction.java
VRohach: I double-checked and found that it was a really specific case. Thank you for this comment!

PR #246, opened 2020-03-03
MODSOURMAN-260 Added query param and condition for choosing a service

KITSViktorSkibnev:
In scope of story MODSOURMAN-260 I added query param defaultMapping and condition for choosing a service depending on query param value.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,8 +7,7 @@
 * Added mode of issuance mapping mechanism
 * Updated mapping for instance identifier types and unspecified instance type
 * Added JobExecutionProgress service
-* Added defaultMapping query param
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -7,6 +7,8 @@
 * Added mode of issuance mapping mechanism
 * Updated mapping for instance identifier types and unspecified instance type
 * Added JobExecutionProgress service
+* Added defaultMapping query param
+* Added condition for choosing of ChunkProcessingService implementation
```
KaterynaSenchenko: Try to combine in one line, smth like
```suggestion
*  Added defaultMapping query param to choose between default mapping and application of JobProfiles
```

PR #247, opened 2020-03-04
MODSOURMAN-241:Implement endpoint to get event with saved instance. 

VRohach:
JIRA-ticket: https://issues.folio.org/browse/MODSOURMAN-241


RuslanLavrov:
update please news

PR #249, opened 2020-03-06
MODSOURMAN-243 - Implement endpoint to receive event about processed record

RuslanLavrov:


mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerHandlersImpl.java
```diff
@@ -44,26 +50,21 @@ public void postChangeManagerHandlersCreatedInventoryInstance(String entity, Map
...
-        // endpoint will be implemented in https://issues.folio.org/browse/MODSOURMAN-244
-        LOGGER.info("Event was received: {}", entity);
-        Future.succeededFuture((Response) ChangeManagerHandlers.PostChangeManagerHandlersDataImportErrorResponse.respond200())
-          .setHandler(asyncResultHandler);
-      } catch (Exception e) {
-        LOGGER.error("Failed to handle event", e);
-        asyncResultHandler.handle(Future.succeededFuture(ExceptionHelper.mapExceptionToResponse(e)));
-      }
+        LOGGER.debug("Event was received: {}", entity);
+        asyncResultHandler.handle(Future.succeededFuture(ChangeManagerHandlers.PostChangeManagerHandlersProcessingResultResponse.respond200()));
```
KaterynaSenchenko: why not 204?

mod-source-record-manager-server/src/main/java/org/folio/services/EventDrivenChunkProcessingServiceImpl.java
```diff
@@ -154,4 +161,31 @@ private boolean isRecordReadyToSend(Record record) {
...
+        if (isAllChunksError) {
+          StatusDto statusDto = new StatusDto().withStatus(StatusDto.Status.ERROR).withErrorStatus(StatusDto.ErrorStatus.RECORD_UPDATE_ERROR);
+          return jobExecutionProgressService.getByJobExecutionId(jobExecutionId, params.getTenantId())
+            .compose(progress -> updateJobExecutionState(jobExecutionId, progress, statusDto, params));
+        }
+        return Future.succeededFuture(false);
+      });
+  }
+
+  private Future<Boolean> updateJobExecutionState(String jobExecutionId, JobExecutionProgress progress, StatusDto statusDto, OkapiConnectionParams params) {
```
KaterynaSenchenko: don't we have such method in JobExecutionService or smth?

mod-source-record-manager-server/src/main/java/org/folio/services/RecordProcessedEventHandlingServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+        });
+    } catch (Exception e) {
+      LOGGER.error("Failed to handle event {}", e, eventContent);
+      promise.fail(e);
+    }
+    return promise.future();
+  }
+
+  private JobExecutionProgress changeProgressAccordingToEventType(JobExecutionProgress progress, DataImportEventTypes eventType) {
+    switch (eventType) {
```
VRohach: What if there won`t be DI_COMPLETED or DI_ERROR? Do we need the default clause?

mod-source-record-manager-server/src/main/java/org/folio/services/RecordProcessedEventHandlingServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+              .withUiStatus(JobExecution.UiStatus.fromValue(Status.valueOf(statusToUpdate.name()).getUiStatus()))
+              .withCompletedDate(new Date())
+              .withProgress(new Progress()
+                .withJobExecutionId(jobExecutionId)
+                .withCurrent(progress.getCurrentlySucceeded() + progress.getCurrentlyFailed())
+                .withTotal(progress.getTotal()));
+
+            return jobExecutionService.updateJobExecutionWithSnapshotStatus(jobExecution, params).map(true);
+          })
+          .orElse(Future.failedFuture(format("Couldn't find JobExecution for update status and progress with id '%s'", jobExecutionId))));
```
VRohach: I think it will be better to change "Couldn't" on "Can't"

mod-source-record-manager-server/src/main/java/org/folio/services/EventDrivenChunkProcessingServiceImpl.java
RuslanLavrov: no, we have not

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerHandlersImpl.java
RuslanLavrov: changed to 204

PR #278, opened 2020-05-01
MODSOURMAN-276 Created endpoint for retrieving MARC record by instance id

KaterynaSenchenko:


mod-source-record-manager-server/src/main/java/org/folio/services/ParsedRecordServiceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public Future<ParsedRecordDto> getRecordByInstanceId(String instanceId, OkapiConnectionParams params) {
+    Promise<ParsedRecordDto> promise = Promise.promise();
+    SourceStorageClient client = new SourceStorageClient(params.getOkapiUrl(), params.getTenantId(), params.getToken());
+    try {
+      client.getSourceStorageSourceRecordsById(instanceId, "INSTANCE", response -> {
+        if (HTTP_OK.toInt() == response.statusCode()) {
+          response.bodyHandler(body -> promise.handle(Try.itGet(() -> mapSourceRecordToParsedRecordDto(body))));
+        } else {
```
RuslanLavrov, line 35: I propose to respond 404 in case we get 404 from SRS. What do you think @KaterynaSenchenko

KaterynaSenchenko: was too lazy to do it, but OK, added

RuslanLavrov: Thank you)!

PR #281, opened 2020-05-05
MODSOURMAN-305: Apply archive/unarchive eventPayload mechanism

VRohach:
1. Zipping and unzipping event payload added to specific places.
2. Tests added.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/RecordProcessedEventHandlingServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/RecordProcessedEventHandlingServiceImpl.java
@@ -17,6 +17,7 @@ import org.folio.services.progress.JobExecutionProgressService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.io.IOException;
 import java.util.Date;
 
 import static java.lang.String.format;
@@ -40,9 +41,17 @@ public class RecordProcessedEventHandlingServiceImpl implements EventHandlingSer
   @Override
   public Future<Boolean> handle(String eventContent, OkapiConnectionParams params) {
     Promise<Boolean> promise = Promise.promise();
+    DataImportEventPayload dataImportEventPayload;
     try {
-      DataImportEventPayload dataImportEventPayload = ObjectMapperTool.getMapper().readValue(ZIPArchiver.unzip(eventContent), DataImportEventPayload.class);
-      String jobExecutionId = dataImportEventPayload.getJobExecutionId();
+      dataImportEventPayload = ObjectMapperTool.getMapper().readValue(ZIPArchiver.unzip(eventContent), DataImportEventPayload.class);
+    } catch (IOException e) {
+      LOGGER.error("Failed to unzip event {}", e, eventContent);
+      promise.fail(e);
+      return promise.future();
+    }
+
+    String jobExecutionId = dataImportEventPayload.getJobExecutionId();
+    try {
       DataImportEventTypes eventType = DataImportEventTypes.valueOf(dataImportEventPayload.getEventType());
       jobExecutionProgressService.updateJobExecutionProgress(jobExecutionId, progress -> changeProgressAccordingToEventType(progress, eventType), params.getTenantId())
         .compose(updatedProgress -> updateJobExecutionIfAllRecordsProcessed(jobExecutionId, updatedProgress, params))
@@ -58,6 +67,7 @@ public class RecordProcessedEventHandlingServiceImpl implements EventHandlingSer
         });
     } catch (Exception e) {
       LOGGER.error("Failed to handle event {}", e, eventContent);
+      updateJobStatusToError(jobExecutionId, params);
       promise.fail(e);
     }
     return promise.future();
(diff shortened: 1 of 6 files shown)


VRohach:
This PR can be merged only after https://github.com/folio-org/data-import-processing-core/pull/45.

mod-source-record-manager-server/src/main/java/org/folio/services/RecordProcessedEventHandlingServiceImpl.java
```diff
@@ -40,7 +41,7 @@ public RecordProcessedEventHandlingServiceImpl(@Autowired JobExecutionProgressSe
   public Future<Boolean> handle(String eventContent, OkapiConnectionParams params) {
     Promise<Boolean> promise = Promise.promise();
     try {
-      DataImportEventPayload dataImportEventPayload = ObjectMapperTool.getMapper().readValue(eventContent, DataImportEventPayload.class);
+      DataImportEventPayload dataImportEventPayload = ObjectMapperTool.getMapper().readValue(ZIPArchiver.unzip(eventContent), DataImportEventPayload.class);
```
OleksiiKuzminov, line 44: Update job status in catch section plz in case of error

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -39,45 +42,50 @@ private EventHandlingUtil() {
...
+        .whenComplete((ar, throwable) -> {
+          if (throwable == null) {
+            promise.complete(record);
+          } else {
+            LOGGER.error("Error during event sending: {}", throwable, createdRecordEvent);
+            promise.fail(throwable);
+          }
+        });
+    } catch (Exception e) {
+      LOGGER.error("Error during event building or sending: {}", e);
```
KaterynaSenchenko, line 86:
```suggestion
      LOGGER.error("Error during event building or sending", e);
```
no need to log exception as a param

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -39,45 +42,50 @@ private EventHandlingUtil() {
...
-          LOGGER.error("Error during event sending: {}", throwable, createdRecordEvent);
-          promise.fail(throwable);
-        }
-      });
+      PubSubClientUtils.sendEventMessage(createdRecordEvent, connectionParams)
+        .whenComplete((ar, throwable) -> {
+          if (throwable == null) {
+            promise.complete(record);
+          } else {
+            LOGGER.error("Error during event sending: {}", throwable, createdRecordEvent);
```
KaterynaSenchenko, line 81: I don't think it is necessary to log the whole createdRecordEvent entity, id and/or event type should suffice

mod-source-record-manager-server/src/test/java/org/folio/services/EventDrivenChunkProcessingServiceImplTest.java
```diff
@@ -165,7 +171,12 @@ public void shouldProcessChunkOfRawRecords(TestContext context) {
       List<LoggedRequest> loggedRequests = findAll(postRequestedFor(urlEqualTo(PUBSUB_PUBLISH_URL)));
       context.assertEquals(1, loggedRequests.size());
       Event event = Json.decodeValue(loggedRequests.get(0).getBodyAsString(), Event.class);
-      DataImportEventPayload dataImportEventPayload = Json.decodeValue(event.getEventPayload(), DataImportEventPayload.class);
+      DataImportEventPayload dataImportEventPayload = null;
+      try {
+        dataImportEventPayload = Json.decodeValue(ZIPArchiver.unzip(event.getEventPayload()), DataImportEventPayload.class);
+      } catch (IOException e) {
```
KaterynaSenchenko, line 177: why not just throw this exception in the method signature if nothing done when it's caught?

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
VRohach: Sure, thanks

mod-source-record-manager-server/src/test/java/org/folio/services/EventDrivenChunkProcessingServiceImplTest.java
VRohach: You are right

VRohach: @KaterynaSenchenko It is needed because this exception will be in the handler, and there aren`t any ways to handle it in the method level.

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
VRohach: @KaterynaSenchenko Good point! But I think, in case of error it will be better to get the whole information about the event. @OleksiiKuzminov what do you think about it?

PR #288, opened 2020-05-26
`MODSOURMAN-275: Remove preview area's sample data.

VRohach:
1. "setup_test_data.sql" removed.
2. Sample data setuping removed from ModTenantAPI.java.`


RuslanLavrov:
guys, do we need a data migration script to remove samples?

PR #289, opened 2020-05-27
MODSOURMAN-318 - Remove hardcoded diku tenant in db schema.json

RuslanLavrov:


mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -89,7 +89,7 @@
     },
     {
       "run": "after",
-      "snippet": "DELETE FROM diku_mod_source_record_manager.job_execution_progress jep1 USING diku_mod_source_record_manager.job_execution_progress jep2 WHERE  jep1.ctid < jep2.ctid AND jep1.jobexecutionid = jep2.jobexecutionid;",
+      "snippet": "DELETE FROM job_execution_progress jep1 USING diku_mod_source_record_manager.job_execution_progress jep2 WHERE  jep1.ctid < jep2.ctid AND jep1.jobexecutionid = jep2.jobexecutionid;",
```
hjiebsco, line 92: Is the "diku" in jep2 by design? Can you explain the intent of this deletion sql?

RuslanLavrov: "diku" in jep2 isn't by design, I'll fix it now. Script intended to delete duplicated progress.

hjiebsco: > "diku" in jep2 isn't by design, I'll fix it now. Script intended to delete duplicated progress. Thanks.

I see. Would unique index on jobExecutionId field help?

RuslanLavrov: Yes, we already added it.

PR #304, opened 2020-07-03
MODSOURMAN-324: Hardcode JobProfile for importing records using CLI tool.

VRohach:
1. Name to jobProfileInfo was added as "CLI job profile".
2. NEWS updated.


mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -237,8 +237,10 @@
   private List<JobExecution> prepareJobExecutionList(String parentJobExecutionId, List<File> files, UserInfo userInfo, InitJobExecutionsRqDto dto) {
     String userId = dto.getUserId();
     if (dto.getSourceType().equals(InitJobExecutionsRqDto.SourceType.ONLINE)) {
+      JobProfileInfo jobProfileInfo = dto.getJobProfileInfo();
+      jobProfileInfo.withName("CLI job profile");
```
KaterynaSenchenko, line 241: Discuss the name with AM

VRohach: @KaterynaSenchenko Done. I changed this name.

PR #306, opened 2020-07-07
MODDATAIMP-309 - Data Import stops when trying to load a large file to folio-snapshot-load

RuslanLavrov:


mod-source-record-manager-server/src/test/java/org/folio/services/JobExecutionProgressServiceImplTest.java
```diff
@@ -102,14 +104,15 @@ public void shouldNotReturnFailedFutureWhenJobExecutionIdDuplicates(TestContext
     Async async = context.async();
     int expectedTotalRecords = 62;
 
-    Future<JobExecutionProgress> future = jobExecutionService.initializeJobExecutions(initJobExecutionsRqDto, params)
-      .compose(initJobExecutionsRsDto -> jobExecutionProgressService.initializeJobExecutionProgress(initJobExecutionsRsDto.getParentJobExecutionId(), expectedTotalRecords, TENANT_ID)
-        .compose(jobExecutionProgress -> jobExecutionProgressService.initializeJobExecutionProgress(jobExecutionProgress.getJobExecutionId(), expectedTotalRecords, TENANT_ID)));
+    Future<CompositeFuture> future = jobExecutionService.initializeJobExecutions(initJobExecutionsRqDto, params)
+      .compose(initJobExecutionsRsDto -> {
+        Future<JobExecutionProgress> future1 = jobExecutionProgressService.initializeJobExecutionProgress(initJobExecutionsRsDto.getParentJobExecutionId(), expectedTotalRecords, TENANT_ID);
+        Future<JobExecutionProgress> future2 = jobExecutionProgressService.initializeJobExecutionProgress(initJobExecutionsRsDto.getParentJobExecutionId(), expectedTotalRecords, TENANT_ID);
```
KaterynaSenchenko, line 110: aren't they identical?

KaterynaSenchenko, line 110:
```suggestion
        Future<JobExecutionProgress> future2 = jobExecutionProgressService.initializeJobExecutionProgress(initJobExecutionsRsDto.getJobExecutionId(), expectedTotalRecords, TENANT_ID);
```

RuslanLavrov: yes, they are identical, two concurrent attempts to create progress with the same jobExecution id

RuslanLavrov: initJobExecutionsRsDto doesn’t have getJobExecutionId() method

KaterynaSenchenko: oops, you're right, missed that it's a test

PR #311, opened 2020-07-22
MODSOURMAN-322 Add source-record states storing mechanism for QM edit workflow

OleksiiKuzminov:


mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -93,9 +108,14 @@
       "fromModuleVersion": "mod-source-record-manager-2.2.0"
     },
     {
+      "run": "after",
+      "snippet": "ALTER TABLE source_records_state DROP CONSTRAINT IF EXISTS source_records_state_id_unique; ALTER TABLE source_records_state ADD CONSTRAINT source_records_state_id_unique UNIQUE (sourceRecordId);",
+      "fromModuleVersion": "mod-source-record-manager-2.4.0"
```
KaterynaSenchenko, line 113: it's going to be included in a hotfix, so it should be 2.3.2

mod-source-record-manager-server/src/main/java/org/folio/services/ParsedRecordServiceImpl.java
```diff
@@ -90,7 +104,8 @@ private ParsedRecordDto mapSourceRecordToParsedRecordDto(Buffer body) {
       .withRecordType(ParsedRecordDto.RecordType.fromValue(sourceRecord.getRecordType().value()))
       .withExternalIdsHolder(sourceRecord.getExternalIdsHolder())
       .withAdditionalInfo(sourceRecord.getAdditionalInfo())
-      .withMetadata(sourceRecord.getMetadata());
+      .withMetadata(sourceRecord.getMetadata())
+      .withRecordState(ParsedRecordDto.RecordState.ACTUAL);
```
KaterynaSenchenko, line 108: How would QM guys know about the "actual" state, the one that is being tracked?

PR #322, opened 2020-08-19
MODSOURCE-178 Move 001 value to 035 field before mapping to Instance

KaterynaSenchenko:


KaterynaSenchenko:
Taking the liberty to merge this PR, will create another one in case any changes are requested

PR #353, opened 2020-12-02
MODSOURMAN-380: Expand journalRecord entity with "title" property

VRohach:
1. Script version changed(downgraded) for applying these changes to the current version.

### DIFF
--- a/mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
+++ b/mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
@@ -111,7 +111,7 @@
     {
       "run": "after",
       "snippet": "ALTER TABLE journal_records ADD COLUMN IF NOT EXISTS title text NULL;",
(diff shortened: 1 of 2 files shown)


mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -111,7 +111,7 @@
     {
       "run": "after",
       "snippet": "ALTER TABLE journal_records ADD COLUMN IF NOT EXISTS title text NULL;",
-      "fromModuleVersion": "mod-source-record-manager-2.5.0"
+      "fromModuleVersion": "mod-source-record-manager-2.4.0"
```
RuslanLavrov, line 114: We don't need to decrease the `fromModuleVersion`. Please, leave the version as is, it should be 2.5.0.

VRohach: @RuslanLavrov Thank you! I`ll close this one.

PR #354, opened 2020-12-02
MODSOURMAN-385: Enable OCLC update processing.

VRohach:
1. Additional check for Status=COMMITED added. It will resolve an issue regarding stucking job on 100% when running data-import via CLI.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
@@ -8,6 +8,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
+
 import org.apache.commons.io.FilenameUtils;
 import org.folio.HttpStatus;
 import org.folio.dao.JobExecutionDao;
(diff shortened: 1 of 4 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -240,7 +242,9 @@
     String userId = dto.getUserId();
     if (dto.getSourceType().equals(InitJobExecutionsRqDto.SourceType.ONLINE)) {
       JobProfileInfo jobProfileInfo = dto.getJobProfileInfo();
-      jobProfileInfo.withName(DEFAULT_JOB_PROFILE);
+      if (jobProfileInfo.getId().equals(DEFAULT_JOB_PROFILE_ID)) {
```
KaterynaSenchenko, line 245: you can add this condition to the one above

VRohach: @KaterynaSenchenko  yes, but in that case(which you described), 248-250 lines won`t reachable if it will be only ONLINE SourceType. Is it acceptable for us?

KaterynaSenchenko: Oh, I see, disregard the comment then

OleksiiKuzminov (approved):

RuslanLavrov (approved):

PR #355, opened 2020-12-09
MODSOURMAN-381 - Add endpoint to retrieve a list of jobLogEntryDto

RuslanLavrov:


ramls/metadata-provider.raml
```diff
@@ -85,3 +87,30 @@ resourceTypes:
...
+        sortBy:
+          description: "sorting by field: source_record_order, title, source_record_action_status, instance_action_status, holdings_action_status, item_action_status, order_action_status, invoice_action_status, error"
+          example: source_record_order
+          type: string
+          required: false
+          default: source_record_order
+        order:
+          description: "sorting direction"
+          example: desc
+          enum: [asc, desc]
```
KaterynaSenchenko, line 104: is that the only place when we validate these param values?

RuslanLavrov: yes

KaterynaSenchenko: found it, but thanks for the comprehensive reply)

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -161,4 +174,33 @@ private String prepareSortingClause(String sortBy, String order) {
     return format(ORDER_BY_PATTERN, sortBy, order);
   }
 
+  private JobLogEntryDtoCollection mapRowSetToJobLogDtoCollection(RowSet<Row> rowSet) {
+    JobLogEntryDtoCollection jobLogEntryDtoCollection = new JobLogEntryDtoCollection()
+      .withTotalRecords(0);
+
+    rowSet.forEach(row -> {
+      JobLogEntryDto jobLogEntryDto = new JobLogEntryDto()
+        .withJobExecutionId(row.getValue("job_execution_id").toString())
```
VRohach, line 183: I think it would be better to extract these fields in constants("source_record_order", "title" and others). Because they are repeating in the default sets too. It will help to avoid typo errors.

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -161,4 +174,33 @@ private String prepareSortingClause(String sortBy, String order) {
...
+
+      jobLogEntryDtoCollection
+        .withTotalRecords(row.getInteger("total_count"))
+        .getEntries().add(jobLogEntryDto);
+    });
+    return jobLogEntryDtoCollection;
+  }
+
+  private org.folio.rest.jaxrs.model.ActionStatus mapNameToEntityActionStatus(String name) {
+    return name == null ? null : org.folio.rest.jaxrs.model.ActionStatus.fromValue(name);
```
VRohach, line 203: @RuslanLavrov can there be the case when "name" will be NOT null, but EMPTY? If yes, then it would be better to use isBlank()-method.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
```diff
@@ -83,4 +87,22 @@ public void getMetadataProviderJournalRecordsByJobExecutionId(String jobExecutio
...
+
+    vertxContext.runOnContext(v -> {
+      try {
+        journalRecordService.getJobLogEntryDtoCollection(jobExecutionId, sortBy, order.name(), limit, offset, tenantId)
+          .map(GetMetadataProviderJobLogEntriesByJobExecutionIdResponse::respond200WithApplicationJson)
+          .map(Response.class::cast)
+          .otherwise(ExceptionHelper::mapExceptionToResponse)
+          .onComplete(asyncResultHandler);
+      } catch (Exception e) {
+        asyncResultHandler.handle(Future.succeededFuture(ExceptionHelper.mapExceptionToResponse(e)));
```
VRohach, line 102: Let`s log smth about the error here.

ramls/metadata-provider.raml
VRohach: @RuslanLavrov @KaterynaSenchenko But there is additional validation for "sortBy" parameter in JournalRecordDaoImpl:
`    if (sortBy != null && !jobLogEntrySortableFields.contains(sortBy)) {`

So, if it's needed, @RuslanLavrov could you please add validation for "order" also. Thank you!

RuslanLavrov: @VRohach, since `sortBy` has the default value and can not be null, sortBy != null validation rather is redundant, hence I'll remove it, thanks

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
RuslanLavrov: @VRohach, do you propose to move column names to constants that appear only once as well?

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
RuslanLavrov: no

VRohach: @RuslanLavrov  Ok, thank you!

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
VRohach: @RuslanLavrov  If there are some properties that appear once, I think it would be a better move to constants them also, for the same code style. Moreover, they can be used in DAO in the future (for example in the next endpoint). What do you think?

RuslanLavrov: I'll move them

VRohach (approved):

OleksiiKuzminov (approved):

PR #358, opened 2020-12-11
MODSOURMAN-382 - Fill title from marc record into journalRecord entity

RuslanLavrov:


mod-source-record-manager-server/src/main/java/org/folio/services/util/ParsedRecordUtil.java
```diff
@@ -0,0 +1,36 @@
...
+import static org.apache.commons.lang3.StringUtils.SPACE;
+
+public final class ParsedRecordUtil {
+
+  public static final String SUBFIELDS_DATA_DELIMITER = SPACE;
+
+  private ParsedRecordUtil() {
+  }
+
+  public static String retrieveDataByField(ParsedRecord parsedRecord, String fieldTag, List<String> subfieldCodes) {
```
VRohach, line 29: Let`s add some docs for this Util/method.

mod-source-record-manager-server/src/main/java/org/folio/services/MappingRuleCache.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private MappingRuleDao mappingRuleDao;
+  private AsyncLoadingCache<String, Optional<JsonObject>> cache;
+
+  @Autowired
+  public MappingRuleCache(MappingRuleDao mappingRuleDao, Vertx vertx) {
+    this.mappingRuleDao = mappingRuleDao;
+    cache = Caffeine.newBuilder()
+      .executor(task -> vertx.runOnContext(ar -> task.run()))
+      .maximumSize(1)
```
VRohach, line 32: I think it would be better to extract this number(1) to the property(field in this class). For better reading and fixing some issues in the future.

mod-source-record-manager-server/src/main/java/org/folio/services/MappingRuleCache.java
```diff
@@ -0,0 +1,65 @@
...
+import org.folio.dao.MappingRuleDao;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+
+@Component
+public class MappingRuleCache {
```
VRohach, line 20: Let`s add some docs here for methods for better understanding this mechanism.

VRohach (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/util/ParsedRecordUtil.java
RuslanLavrov: it has already been added in the previous commit)

mod-source-record-manager-server/src/main/java/org/folio/services/MappingRuleCache.java
RuslanLavrov: missed it, it should be deleted since cache can stores rules for multiple tenants, thanks

mod-source-record-manager-server/src/main/java/org/folio/services/MappingRuleCache.java
RuslanLavrov: done

OleksiiKuzminov (approved):

PR #359, opened 2020-12-14
MODSOURMAN-383:Implement endpoint to retrieve a recordProcessingLogDto.

VRohach:
1. Created function for retrieving data needed for recordProcessingLogDto. Function`s name: "get_record_processing_log"
2.  Created endpoint for retrieving recordProcessingLogDto. URL: **/jobLogEntries/{jobExecutionId}/records/{recordId}**
3. Added new methods for getting this data to the JournalRecordDao and JournalRecordService.
4. Added mechanism for constructing recordProcessingLogDto from script`s result.
5. NEWS updated.
6. Tests were added.


VRohach:
Tests were added.
So, @OleksiiKuzminov and @RuslanLavrov, feel free to review!

descriptors/ModuleDescriptor-template.json
```diff
@@ -281,6 +281,17 @@
         }
       ]
     },
+    {
+      "id": "metadata-provider-record-processing-logs",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/metadata-provider/jobLogEntries/{jobExecutionId}/records/{recordId}",
```
RuslanLavrov, line 290: propose that we move this endpoint/handler to the previous `metadata-provider-job-execution-logs` interface, which will be responsible for providing data for logs lite

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDao.java
```diff
@@ -62,4 +63,14 @@
    * @return future with JobLogEntryDto collection
    */
   Future<JobLogEntryDtoCollection> getJobLogEntryDtoCollection(String jobExecutionId, String sortBy, String order, int limit, int offset, String tenantId);
+
+  /**
+   * Searches for JobLogEntryDto entities by jobExecutionId and recordId
```
RuslanLavrov, line 68:
```suggestion
   *Searches for RecordProcessingLogDto entity by jobExecutionId and recordId
```

mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
```diff
@@ -0,0 +1,99 @@
+-- Script to create function to get data import record processing log entries (recordProcessingLogDto).
+CREATE OR REPLACE FUNCTION get_record_processing_log(jobExecutionId uuid, recordId uuid)
+    RETURNS TABLE(job_execution_id uuid, source_id uuid, source_record_order integer, title text, source_record_action_status text, source_entity_id uuid, source_entity_hrid text, source_entity_error text, instance_action_status text, instance_entity_id uuid, instance_entity_hrid text, instance_entity_error text, holdings_action_status text, holdings_entity_id uuid, holdings_entity_hrid text, holdings_entity_error text, item_action_status text, item_entity_id uuid, item_entity_hrid text, item_entity_error text, order_action_status text, order_entity_id uuid, order_entity_hrid text, order_entity_error text, invoice_action_status text, invoice_entity_id uuid, invoice_entity_hrid text, invoice_entity_error text)
+AS $$
+BEGIN
+    RETURN QUERY EXECUTE format('
```
RuslanLavrov, line 6:
```suggestion
    RETURN QUERY
```
it looks like we can do without dynamic command, hence here we can avoid redundant format() function call and directly write a query.

mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
```diff
@@ -0,0 +1,99 @@
...
+			   invoice_entity_hrid[1],
+			   invoice_entity_error[1]
+        FROM (
+               SELECT journal_records.source_id,
+                      journal_records.source_record_order,
+                      journal_records.job_execution_id,
+                      array_agg(action_type ORDER BY action_date ASC) FILTER (WHERE entity_type = ''MARC_BIBLIOGRAPHIC'') AS marc_actions,
+                      count(journal_records.source_id) FILTER (WHERE entity_type = ''MARC_BIBLIOGRAPHIC'' AND journal_records.error != '''') AS marc_errors_number,
+
+			          array_agg(entity_hrid) FILTER (WHERE entity_type = ''MARC_BIBLIOGRAPHIC'') AS source_entity_hrid,
```
RuslanLavrov, line 48: let's remove it since a source record has not hrid

mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
```diff
@@ -0,0 +1,99 @@
...
+			          array_agg(entity_id) FILTER (WHERE entity_type = ''MARC_BIBLIOGRAPHIC'') AS source_entity_id,
```
RuslanLavrov, line 49: let's remove this, looks like we don't use it to fill the DTO

mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
```diff
@@ -0,0 +1,99 @@
...
+			   source_entity_id[1]::uuid,
+			   source_entity_hrid[1],
+			   source_entity_error[1],
+               get_entity_status(instance_actions, instance_errors_number) AS instance_action_status,
+			   instance_entity_id[1]::uuid,
+			   instance_entity_hrid[1],
+			   instance_entity_error[1],
+               get_entity_status(holdings_actions, holdings_errors_number) AS holdings_action_status,
+			   holdings_entity_id[1]::uuid,
+			   holdings_entity_hrid[1],
```
RuslanLavrov, line 27: let's return all array elements, it is assumed that multiple entities of the same type can be created/updated by the same source record

RuslanLavrov (approved):

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/metadataProvider/MetaDataProviderJobLogEntriesAPITest.java
```diff
@@ -314,7 +317,173 @@ public void shouldReturnLimitedCollectionOnGetWithLimitAndOffset(TestContext con
...
+
+
+  @Test
+  public void shouldReturnEmptyDtoIfHasNoLogRecordsBySpecifiedJobIdAndRecordId() {
+    RestAssured.given()
+      .spec(spec)
+      .when()
+      .get(GET_JOB_EXECUTION_JOURNAL_RECORDS_PATH + "/" + UUID.randomUUID().toString() + "/records/" + UUID.randomUUID().toString())
+      .then()
+      .statusCode(HttpStatus.SC_OK);
```
RuslanLavrov, line 329: maybe it would better to return 404 status code instead of empty object when has no information for specified jobId and recordId

mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
VRohach: @RuslanLavrov As far as  I understand, we have this field in the schema: "sourceRecordId". Am I wrong?

RuslanLavrov: @VRohach, yes agree. As I see we use "source_id" column to fill "sourceRecordId" property.  Since we have record id in "source_id" column I propose to remove this calculation.

VRohach: @RuslanLavrov Ok. Got it! Thank you

OleksiiKuzminov (approved):

PR #364, opened 2021-01-22
MODSOURMAN-399: Edifact parser

jeremythuff:
## Purpose

This PR satisfies the requirements of [MODSOURMAN-399](https://issues.folio.org/browse/MODSOURMAN-399), allowing for the parsing of raw EDIFACT into a standardized json representation.

## Approach

This was accomplished by adding the `EdifactRecordParser` implementation of the `RecordParser` interface. 

requires https://github.com/folio-org/data-import-raml-storage/pull/197


mod-source-record-manager-server/src/main/java/org/folio/services/parsers/EdifactRecordParser.java
```diff
@@ -0,0 +1,146 @@
...
+import io.vertx.core.logging.LoggerFactory;
+import io.xlate.edi.stream.EDIInputFactory;
+import io.xlate.edi.stream.EDIStreamReader;
+import io.xlate.edi.stream.EDIStreamValidationError;
+
+/**
+ * Raw record parser implementation for EDIFACT format. Use staedi library
+ */
+public final class EdifactRecordParser implements RecordParser {
+  private static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
```
OleksiiKuzminov, line 26: As I know, not all files can be in UTF-8

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -333,6 +337,16 @@ public static boolean isStatus(HttpClientResponse response, HttpStatus status) {
     return journalRecords;
   }
 
+  private JournalRecord.EntityType inferJournalRecordEntityType(Record record) {
+    switch (record.getRecordType()) {
+      case EDIFACT:
+        return JournalRecord.EntityType.INVOICE;
```
RuslanLavrov, line 343: Hello @jeremythuff, could you please expand [the schema of `EntityType` enum](https://github.com/folio-org/data-import-raml-storage/blob/e48acdc3f446dd40ba7d9c4743b392f8a4b7a757/schemas/mod-source-record-manager/journalRecord.json#L24) with `EDIFACT` value to indicate that EDIFACT record was created

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
jeremythuff: yes I can

RuslanLavrov: thank you!

wwelling (approved):

PR #369, opened 2021-02-01
Fix multiple completion of the same future

KaterynaSenchenko:
## Purpose
ConsumersVerticle start and stop futures complete multiple times

## Approach
Use Composite future

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
@@ -1,6 +1,8 @@
 package org.folio.verticle;
 
 import io.vertx.core.AbstractVerticle;
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
 import io.vertx.core.Promise;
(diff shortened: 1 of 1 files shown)


RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
```diff
@@ -54,17 +56,20 @@ public void start(Promise<Void> startPromise) {
...
   @Override
   public void stop(Promise<Void> stopPromise) {
-    consumerWrappersList.forEach(consumerWrapper -> consumerWrapper
-      .stop()
-      .onComplete(ar -> stopPromise.complete()));
+    List<Future> futures = new ArrayList<>();
+    consumerWrappersList.forEach(consumerWrapper ->
+      futures.add(consumerWrapper.stop()));
+
+    CompositeFuture.all(futures).onComplete(ar -> stopPromise.complete());
```
RuslanLavrov, line 72:
```suggestion
    CompositeFuture.join(futures).onComplete(ar -> stopPromise.complete());
```
suggest to make sure that all consumers have been stopped.
Thank you, Kate!

PR #398, opened 2021-03-24
MODSOURMAN-407: Create JobExecution seems to require jobProfileInfo.

VRohach:
## Purpose
Remove validation for JobProfileInfo if JobExecution is being created with ONLINE param.

## Approach
Remove validation for JobProfileInfo if JobExecution is being created with ONLINE param
Add tests

## Learning
More info: https://issues.folio.org/browse/MODSOURMAN-407

**Note:** moreover, an additional check on user's personal data was added. It can avoid NullPointerException in cases when the user doesn't have any personal data. Moreover logging for this case was added.


KaterynaSenchenko (commented):
please check documentation - step when online processing is initiated, and update example request if needed

VRohach:
> please check documentation - step when online processing is initiated, and update example request if needed

Done. Could you please verify ?

KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

PR #420, opened 2021-04-26
MODSOURMAN-453:  Add index for the " job_execution_source_chunks"

VRohach:
## Purpose
Increase performance and remove extra logging

## Approach
Full-text search changed on equals search.
Add 2 indices.

## Learning
More info: https://issues.folio.org/browse/MODSOURMAN-453


RuslanLavrov (approved):

mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -2,7 +2,7 @@
   "tables": [
     {
       "tableName": "job_executions",
-      "fromModuleVersion": "mod-source-record-manager-3.0.5",
+      "fromModuleVersion": "mod-source-record-manager-3.0.6",
```
KaterynaSenchenko, line 5: By default it won't be released as a hotfix, please use the 3.1.0 version. If it is crucial for Iris, then we'll change it

mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -26,7 +29,7 @@
     },
     {
       "tableName": "job_execution_source_chunks",
-      "fromModuleVersion": "mod-source-record-manager-1.2.0",
+      "fromModuleVersion": "mod-source-record-manager-3.0.6",
```
KaterynaSenchenko, line 32: Same here

mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
VRohach: Changed

mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
VRohach: Changed

KaterynaSenchenko (approved):

PR #421, opened 2021-05-05
MODSOURCE-278 - Record from chunk is not saved occasionally

RuslanLavrov:
## Purpose
to correctly provide records creation info for the log summary page in case of error while saving records in the record-storage

## Approach
* move logging of the records creation information to the handler receiving saved records
* tests

## Learning
[MODSOURCE-278](https://issues.folio.org/browse/MODSOURCE-278)


KaterynaSenchenko (approved):

KaterynaSenchenko (commented):
please add news

OleksiiKuzminov (approved):

PR #422, opened 2021-05-06
MODSOURMAN-461: Data Import fails (no details about cause of failure in ui/log)

VRohach:
## Purpose
The main goal is to avoid an issue regarding that srs record can`t be edited in quickMarc and can't select "View source" on it.
The reason was that "instanceId" didn't apply for a specific record in the "records_lb" table. That's happened because srs received a record without "externalIdsHolder". That's why "instanceId" is empty in DB.

## Approach
There was added mechanism for setting instanceId to the record(from 999ff$i field), if this is not the first import for this record(it was already created based on matchedId).

## Learning
More info: https://issues.folio.org/browse/MODSOURMAN-461
Cases for reproducing was retrieved from these tickets:  https://issues.folio.org/browse/MODDATAIMP-426 and https://issues.folio.org/browse/MODDATAIMP-427


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -207,6 +208,10 @@ private boolean actionProfileMatches(ProfileSnapshotWrapper actionProfileWrapper
             if (matchedId != null) {
               record.setMatchedId(matchedId);
               record.setGeneration(null); // in case the same record is re-imported, generation should be calculated on SRS side
+              String instanceId = getValue(record, "999", 'i');
+              if (instanceId != null) {
```
OleksiiKuzminov, line 212: can be empty?

VRohach: @OleksiiKuzminov Thank you for this comment!
Added check for both: matchedId(from previous logic) and instanceId.

RuslanLavrov (approved):

PR #425, opened 2021-05-12
MODSOURMAN-471 Migrate QM-flow to Kafka

psmagin:
## Purpose
Migrate QM record update flow to Kafka.

## Approach
* Support consuming `QM_INVENTORY_INSTANCE_UPDATED` and `QM_ERROR` events
* Support producing `QM_RECORD_UPDATED` event
* Deploying new verticle for consumer


natali-zaitseva (approved):

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/InitAPIImpl.java
```diff
@@ -42,6 +43,9 @@
   @Value("${srm.kafka.DataImportJournalConsumersVerticle.instancesNumber:5}")
   private int dataImportJournalConsumerInstancesNumber;
 
+  @Value("${srm.kafka.QuickMarcUpdateConsumersVerticle.instancesNumber:5}")
```
KaterynaSenchenko, line 46: please add this to the readme file

KaterynaSenchenko (approved):

Igor-Gorchakov (approved):

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/InitAPIImpl.java
psmagin: added

PR #452, opened 2021-07-16
MODSOURMAN-517 - Change quickMarc producers not to close

Igor-Gorchakov:
https://issues.folio.org/browse/MODSOURMAN-517

## Purpose
Change quickMarc producers not to close after message sent. Currently, they are created each time the event send and then closed

## Approach
Create shared producers and store them in map
Do not close producer after message sent

#### TODOS and Open Questions
- [ ] Use GitHub checklists. When solved, check the box and explain the answer.

## Learning
_Describe the research stage. Add links to blog posts, patterns, libraries or addons used to solve this problem._

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
@@ -17,13 +17,13 @@ import org.folio.rest.jaxrs.model.EventMetadata;
 import org.folio.util.pubsub.PubSubClientUtils;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
 public final class EventHandlingUtil {
-  private final static ConcurrentMap<String, KafkaProducer<String, String>> KAFKA_PRODUCERS = new ConcurrentHashMap<>();
+  private final static Map<String, KafkaProducer<String, String>> KAFKA_PRODUCERS = new HashMap<>();
 
   private EventHandlingUtil() {
   }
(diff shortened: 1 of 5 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -19,8 +19,11 @@
 import java.io.IOException;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 public final class EventHandlingUtil {
+  private final static ConcurrentMap<String, KafkaProducer<String, String>> KAFKA_PRODUCERS = new ConcurrentHashMap<>();
```
psmagin, line 26: 1. I don't think that using `ConcurrentMap` is a good idea because it will block Vert.X EventBus
2. Do you really want to change the logic of producers for all logic in a module including Data Import features? Maybe will be better to have this change only for quickMarc.

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -89,4 +91,13 @@ private static String extractCorrelationId(List<KafkaHeader> kafkaHeaders) {
...
   }
 
+  private static KafkaProducer<String, String> getKafkaProducer(String producerName, KafkaConfig kafkaConfig) {
+    if (KAFKA_PRODUCERS.containsKey(producerName)) {
+      return KAFKA_PRODUCERS.get(producerName);
+    } else {
+      KafkaProducer<String, String> kafkaProducer = KafkaProducer.createShared(Vertx.currentContext().owner(), producerName, kafkaConfig.getProducerProps());
+      KAFKA_PRODUCERS.put(producerName, kafkaProducer);
+      return kafkaProducer;
+    }
```
psmagin, line 101: Could be replaced with `computeIfAbsent`

psmagin (changes requested):

psmagin (approved):

natali-zaitseva (approved):

PR #468, opened 2021-08-19
MODSOURMAN-533 Upgrade to RAML Module Builder 33.x

bisecomsergiy:
## Purpose
In the scope of the task on module upgrade, proper actions were fulfilled following instructions to upgrade RAML

## Approach
- Enabled generation equals method for model
- Excluded unnecessary clients in the final client jar
- Added commons-lang artifact
- Updated tests

## Learning
https://issues.folio.org/browse/MODSOURMAN-533


SerhiiNosko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -509,7 +509,7 @@ private JobExecution modifyJobExecutionToCompleteWithError(JobExecution jobExecu
         } else {
           String message = format("Records from SRS were not deleted for JobExecution %s", jobExecutionId);
           LOGGER.error(message);
-          promise.fail(new HttpStatusException(response.result().statusCode(), message));
+          promise.fail(new HttpException(response.result().statusCode(), "Error during post for new Snapshot."));
```
VRohach, line 512: @bisecomsergiy Why error message was changed?

bisecomsergiy: Very good point, moved back it. Thank you!

VRohach (approved):

KaterynaSenchenko (approved):

PR #473, opened 2021-08-25
MODSOURMAN-463 Create storage and API for MappingRules and MappingParams

KaterynaSenchenko:
## Purpose
Mapping Rules and Mapping Parameters do not change during processing for a particular JobExecution, therefore they should be removed from DataImportEventPayload to reduce it's size

## Approach
Added DB tables and DAOs to persist MappingRules and MappingParams snapshots for each JobExecution. 
Added new endpoint to provide access to Mapping Metadata, which consists of MappingRules and MappingParams.
Added service to manage MappingRules and MappingParams snapshots.

#### TODOS and Open Questions
- [ ] Tests are WIP
- [x] MappingMetadataDto.json will be removed from this PR once https://github.com/folio-org/data-import-raml-storage/pull/226 is merged


SerhiiNosko (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/MappingParamsSnapshotDaoImpl.java
```diff
@@ -0,0 +1,79 @@
...
+  public Future<Optional<MappingParameters>> getByJobExecutionId(String jobExecutionId, String tenantId) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    String query = format(SELECT_QUERY, convertToPsqlStandard(tenantId), TABLE_NAME);
+    Tuple queryParams = Tuple.of(UUID.fromString(jobExecutionId));
+    pgClientFactory.createInstance(tenantId).select(query, queryParams, promise);
+    return promise.future().map(resultSet -> {
+      if (resultSet.rowCount() == 0) {
+        return Optional.empty();
+      } else {
+        MappingParameters rules = new JsonObject(resultSet.iterator().next().getValue("params").toString()).mapTo(MappingParameters.class);
```
VRohach, line 47: @KaterynaSenchenko  I think there is a potential NullPointerException: `.getValue('params').toString()` 
It would be better to avoid it. What do you think?

mod-source-record-manager-server/src/main/java/org/folio/dao/MappingRulesSnapshotDaoImpl.java
```diff
@@ -0,0 +1,80 @@
...
+  public Future<Optional<JsonObject>> getByJobExecutionId(String jobExecutionId, String tenantId) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    String query = format(SELECT_QUERY, convertToPsqlStandard(tenantId), TABLE_NAME);
+    Tuple queryParams = Tuple.of(UUID.fromString(jobExecutionId));
+    pgClientFactory.createInstance(tenantId).select(query, queryParams, promise);
+    return promise.future().map(resultSet -> {
+      if (resultSet.rowCount() == 0) {
+        return Optional.empty();
+      } else {
+        JsonObject rules = new JsonObject(resultSet.iterator().next().getValue("rules").toString());
```
VRohach, line 45: The same regarding NPE.

mod-source-record-manager-server/src/main/java/org/folio/dao/MappingRulesSnapshotDaoImpl.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Override
+  public Future<String> save(JsonObject rules, String jobExecutionId, String tenantId) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    try {
+      String query = format(INSERT_SQL, convertToPsqlStandard(tenantId), TABLE_NAME);
+      Tuple queryParams = Tuple.of(
+        UUID.fromString(jobExecutionId),
+        rules,
+        LocalDateTime.now()
```
VRohach, line 59: I think it would be better to add timezones, to avoid potential bugs as we already caught for another mechanism. To add timezone we need to make a specific request to another service: https://github.com/folio-org/mod-source-record-manager/blob/483bb7ba358c6daee8d87c6600d94d8f1c902ae9/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/MappingParametersProvider.java#L696

KaterynaSenchenko: we don't care that much for a time when the metadata was saved, it might be useful when we decide to clear the DB of some old irrelevant data, and timezone won't make any difference

RuslanLavrov (approved):

Aliaksandr-Fedasiuk (approved):

PR #475, opened 2021-08-30
MODSOURMAN-547 Update cache for mapping-rules to support MARC Holdings default rules

Shans-Kaluhin:
## Purpose
Update cache that handles mapping rules for MARC Bib to support also MARC Holdings.

## Approach
- Update cache key to have tenantId with record type
- Update all related classes that use cache

## Learning
[MODSOURMAN-547](https://issues.folio.org/browse/MODSOURMAN-547)


mod-source-record-manager-server/src/main/java/org/folio/services/entity/MappingRuleCacheKey.java
```diff
@@ -0,0 +1,66 @@
...
+      this.recordType = Record.RecordType.MARC_BIB;
+    } else if (org.folio.rest.jaxrs.model.Record.RecordType.MARC_HOLDING.equals(recordType)) {
+      this.recordType = Record.RecordType.MARC_HOLDING;
+    } else {
+      throw new BadRequestException(ERROR_MESSAGE);
+    }
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
psmagin, line 55: You can use EqualsAndHashCode Lombok annotation.

psmagin (approved):

natali-zaitseva (approved):

PR #476, opened 2021-09-01
MODSOURMAN-553 Update GET /change-manager/parsedRecords to have externalId param

Shans-Kaluhin:
## Purpose
Update GET /change-manager/parsedRecords  to have externalId param instead instanceId. It will give quick-marc an ability to fetch records by its externalId from SRS.

## Approach
instanceId parameter is replaced with externalId 

## Learning
[MODSOURMAN-553](https://issues.folio.org/browse/MODSOURMAN-553)
[API Documentation](https://s3.amazonaws.com/foliodocs/api/mod-source-record-manager/change-manager.html#change_manager_parsedrecords_get)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -243,7 +243,7 @@
     },
     {
       "id": "source-manager-parsed-records",
-      "version": "2.0",
+      "version": "3.0",
       "handlers": [
         {
           "methods": [
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -446,7 +446,7 @@
     },
     {
       "permissionName": "change-manager.parsedrecords.get",
-      "displayName": "Change Manager - get parsed records by instanceId",
+      "displayName": "Change Manager - get parsed records by externalId",
       "description": "Get parsed record"
     },
     {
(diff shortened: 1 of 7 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -446,7 +446,7 @@
     },
     {
       "permissionName": "change-manager.parsedrecords.get",
-      "displayName": "Change Manager - get parsed records by instanceId",
+      "displayName": "Change Manager - get parsed records by externalId",
```
psmagin, line 449: Update major interface version of interface with this endpoint

psmagin: @natali-zaitseva, what we should do in case of a major version change?

psmagin: Thank you!
As I can see only mod-quick-marc and ui-agreements are using this interface. 
image

natali-zaitseva: and we should it before R3 "Major interface provided to be updated", which seems to be before 17 Sep.

psmagin: I notified ERM team directly about the change. But we still could notify #development channel.

psmagin (approved):

natali-zaitseva (approved):

PR #477, opened 2021-09-08
MODSOURMAN-466 Remove zipping mechanism for data import event payloads

bisecomsergiy:
## Purpose
Within R3 2021 Kiwi Data import Stability/Reliability work, removed zipping mechanism for data import event payloads and tests adjusted

## Approach
Removed zipping mechanisms;
Tests corrected;
QuickMarc zipping mechanisms without change;

## Learning
https://issues.folio.org/browse/MODSOURMAN-466


mod-source-record-manager-server/src/test/java/org/folio/verticle/consumers/RawMarcChunkConsumersVerticleTest.java
```diff
@@ -67,8 +63,6 @@ public static void setUpClass() throws IOException {
   public void setUp() {
     WireMock.stubFor(WireMock.get("/data-import-profiles/jobProfiles/" + JOB_PROFILE_ID + "?withRelations=false&")
       .willReturn(WireMock.ok().withBody(Json.encode(new JobProfile().withId(JOB_PROFILE_ID).withName("Create instance")))));
-    WireMock.stubFor(WireMock.post("/source-storage/batch/verified-records")
```
RuslanLavrov, line 70: @bisecomsergiy, please return back this stub. Thank you!

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
```diff
@@ -111,10 +109,6 @@ public StoredRecordChunksKafkaHandler(@Autowired @Qualifier("recordsPublishingSe
             LOGGER.error("RecordsBatchResponse processing has failed with errors correlationId: {} chunkNumber: {}", correlationId, chunkNumber, th);
             return Future.failedFuture(th);
           });
-      } catch (IOException e) {
```
Aliaksandr-Fedasiuk, line 114: probably it makes sense not to delete try ... catch..., we can change IOException to Exception because we can't catch exceptions from Json.decodeValue (as example)

mod-source-record-manager-server/src/test/java/org/folio/verticle/consumers/RawMarcChunkConsumersVerticleTest.java
bisecomsergiy: Thank you Ruslan, it is done

Aliaksandr-Fedasiuk (approved):

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
bisecomsergiy: Thanks Alex, yes it is discussed with Ruslan and it is done.

SerhiiNosko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -43,7 +43,7 @@ private EventHandlingUtil() {
     LOGGER.debug("Starting to send event to Kafka for eventType: {}", eventType);
     Event event;
     try {
-      event = createEvent(eventPayload, eventType, tenantId, true);
+      event = createEvent(eventPayload, eventType, tenantId, false);
```
VRohach, line 46: @bisecomsergiy  I think that method createEvent should be changed. The last argument should be deleted at all, and zipping inside should be removed as well.

PR #481, opened 2021-09-09
MODSOURMAN-465 Remove MappingRules, MappingParams, and JobProfileSnapshot from the event payload

bisecomsergiy:
## Approach
Under R3 2021 Kiwi Data import Stability/Reliability work:
- MappingRules, MappingParams, and JobProfileSnapshot moved from payloads;
- Tests adjusted;

## Learning
https://issues.folio.org/browse/MODSOURMAN-465


mod-source-record-manager-server/src/main/java/org/folio/services/RecordsPublishingServiceImpl.java
```diff
@@ -165,35 +158,21 @@ private boolean isRecordReadyToSend(Record record) {
...
    * @param params                 connection parameters
    * @return dataImportEventPayload
    */
   private DataImportEventPayload prepareEventPayload(Record record, ProfileSnapshotWrapper profileSnapshotWrapper,
-                                                     JsonObject mappingRules, MappingParameters mappingParameters, OkapiConnectionParams params,
-                                                     String eventType) {
-    HashMap<String, String> context = payloadContextBuilder.buildFrom(record, mappingRules, mappingParameters);
+                                                     OkapiConnectionParams params, String eventType) {
+    HashMap<String, String> context = payloadContextBuilder.buildFrom(record,
+      Objects.isNull(profileSnapshotWrapper) ? StringUtils.EMPTY : profileSnapshotWrapper.getId());
```
SerhiiNosko, line 172: As alternative we can use: `Optional.ofNullable(profileSnapshotWrapper).map(ProfileSnapshotWrapper ::getId).orElse(StringUtils.EMPTY)`.
By the way, maybe better to use null instead of empty string in this case

bisecomsergiy: At the stage, the variable should not be null, so the check is moved. Thank you.

SerhiiNosko (approved):

Aliaksandr-Fedasiuk (approved):

Abdulkhakimov (approved):

AivarIUsupov (approved):

KaterynaSenchenko (approved):

VRohach (approved):

PR #505, opened 2021-10-18
MODSOURMAN-573 - Create mapping rules for AUTHORITY records

Igor-Gorchakov:
## Purpose
The purpose is to have the rules needed for further mapping in mod-inventory

## Approach
 Create marc_authority_rules.json taking mapping rules from https://wiki.folio.org/display/FOLIJET/Search+MARC+Authorities  
The schema is updated accordingly https://github.com/folio-org/mod-inventory/pull/410

#### TODOS and Open Questions
- [ ] Use GitHub checklists. When solved, check the box and explain the answer.

## Learning
_Describe the research stage. Add links to blog posts, patterns, libraries or addons used to solve this problem._

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 ## 2021-xx-xx v3.3.0-SNAPSHOT
 * [MODSOURMAN-577](https://issues.folio.org/browse/MODSOURMAN-577) Optimistic locking: mod-source-record-manager modifications
+* [MODSOURMAN-573](https://issues.folio.org/browse/MODSOURMAN-573) Create mapping rules for AUTHORITY records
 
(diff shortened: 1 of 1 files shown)


mod-source-record-manager-server/src/main/resources/rules/marc_authority_rules.json
```diff
@@ -0,0 +1,738 @@
...
+        "1",
+        "4",
+        "5"
+      ],
+      "rules": []
+    }
+  ],
+  "999": [
+    {
+      "target": "hrid",
```
psmagin, line 730: As I know there should be no hrid in Authority

psmagin (approved):

PR #529, opened 2021-12-03
MODSOURMAN-624 Failed to handle DI_ERROR when 004 is invalid in MARC Holdings

viacheslavkol:
## Purpose
Fix handling DI_ERROR when 004 is invalid in MARC Holdings not to make import stuck

## Approach
- add `recordId` header to kafka event
- cover case with unit-test

## Learning
[MODSOURMAN-624](https://issues.folio.org/browse/MODSOURMAN-624)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -95,6 +95,7 @@ public class ChangeEngineServiceImpl implements ChangeEngineService {
   private static final AtomicInteger indexer = new AtomicInteger();
   private static final String HOLDINGS_004_TAG_ERROR_MESSAGE =
     "The 004 tag of the Holdings doesn't has a link to the Bibliographic record";
+  static final String RECORD_ID = "recordId";
   public static final String MESSAGE_KEY = "message";
 
   private JobExecutionSourceChunkDao jobExecutionSourceChunkDao;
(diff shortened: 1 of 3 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -371,6 +372,8 @@ private void populateError(Record record, JobExecution jobExecution, OkapiConnec
       .withContent(record.getRawRecord().getContent())
       .withDescription(new JsonObject().put(MESSAGE_KEY, HOLDINGS_004_TAG_ERROR_MESSAGE).encode())
     );
+    okapiParams.getHeaders().set(RECORD_ID, record.getId());
+
     sendEventToKafka(okapiParams.getTenantId(), Json.encode(eventPayload), DI_ERROR.value(),
       KafkaHeaderUtils.kafkaHeadersFromMultiMap(okapiParams.getHeaders()), kafkaConfig, key)
       .onFailure(th -> LOGGER.error("Error publishing DI_ERROR event for MARC Holdings record with id {}", record.getId(), th));
```
psmagin, line 379: Will be better to add this header directly to the KafkaHeaders list instead of okapiParams. I think so because `KafkaHeaderUtils.kafkaHeadersFromMultiMap` method could be changed with for example filtering headers by some params. In this case, your change will not work as expected. 
I suggest extracting variable from `KafkaHeaderUtils.kafkaHeadersFromMultiMap(okapiParams.getHeaders())` and add `recordId` header to this collection of Kafka headers.

viacheslavkol: Done in 1ae9a0b

psmagin (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -95,6 +95,7 @@
   private static final AtomicInteger indexer = new AtomicInteger();
   private static final String HOLDINGS_004_TAG_ERROR_MESSAGE =
     "The 004 tag of the Holdings doesn't has a link to the Bibliographic record";
+  static final String RECORD_ID = "recordId";
```
oleksandr-dekin, line 98: Maybe will be better to add modifier?

oleksandr-dekin (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
viacheslavkol: I think package-private is suitable here to make it possible for test to use it and not to expose it globally.

PR #530, opened 2021-12-04
MODSOURMAN-468 - Update API and change db table that should store and represent information for DataImport landing page.

RuslanLavrov:
## Purpose
to create new jobExecutions storage and api to retrieve date for landing page

## Approach
* create new table (non jsonb) for jobExecutions
* change dao/service according to new table
* expose query parameters to support filtering and sorting jobExecutions by necessary criteria for GET /metadata-provider/jobExecutions endpoint
* delete filtering parameter for GET /change-manager/jobExecutions/{parentId}/children endpoint since it supports only jsonb column 
* update/add tests

## Learning
[MODSOURMAN-468](https://issues.folio.org/browse/MODSOURMAN-468)


Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
```diff
@@ -18,13 +21,21 @@
...
+import java.util.Date;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class MetadataProviderImpl implements MetadataProvider {
 
   private static final Logger LOGGER = LogManager.getLogger();
+  private static final Set<Object> JOB_EXECUTION_SORTABLE_FIELDS =
```
AivarIUsupov, line 36: We can use Set of Strings instead of Set of Objects to make it more clear.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -53,111 +67,129 @@
...
+    "WITH cte AS (SELECT count(*) AS total_count FROM %s " +
+    "WHERE subordination_type <> 'PARENT_MULTIPLE' AND %s) " +
+    "SELECT j.*, cte.*, p.jsonb -> 'total' total, " +
+    "(p.jsonb -> 'currentlySucceeded')::int + (p.jsonb -> 'currentlyFailed')::int currently_processed " +
+    "FROM %s j " +
+    "LEFT JOIN %s p ON  j.id = p.jobexecutionid " +
+    "LEFT JOIN cte ON true " +
+    "WHERE subordination_type <> 'PARENT_MULTIPLE' AND %s " +
+    "ORDER BY %s %s " +
+    "LIMIT $1 OFFSET $2";
```
VRohach, line 111: @RuslanLavrov  I think it would be better to extract these SQL queries to the separate final class to not mix up with implementation. Your thoughts?

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionsCache.java
```diff
@@ -40,15 +41,15 @@ public JobExecutionsCache(JobExecutionService jobExecutionService, Vertx vertx,
     cache = buildCache();
   }
 
-  public Future<JobExecutionDtoCollection> get(String tenantId, String cqlQuery, int offset, int limit) {
+  public Future<JobExecutionDtoCollection> get(String tenantId, JobExecutionFilter filter, String sortBy, String order, int offset, int limit) {
```
VRohach, line 44: @RuslanLavrov  Do we have smth like validation for this JobExxecutionFilter?

AivarIUsupov (approved):

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
RuslanLavrov: yes sure, for some reason during extracting to constants type was generated in such way, thanks!

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
RuslanLavrov: done

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionsCache.java
RuslanLavrov: we have validation for Date parameters is that performed by RMB before call our endpoint

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
VRohach: @RuslanLavrov  Thank you!

VRohach (approved):

SerhiiNosko (approved):

PR #540, opened 2021-12-22
MODSOURMAN-653 Authority update: Make Journal to log authorities

Shans-Kaluhin:
### Overview

 To have an ability to Log Authority records that were updated/matched or not matched

### Approach

  Update JournalParams, and DataImportJournalConsumersVerticle

### Acceptance criteria

The matched and non-matched records are getting logged into Journal
The new code is covered by unit tests (JournalParamsTest)

### Learn
https://issues.folio.org/browse/MODSOURMAN-653

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/JournalParams.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/JournalParams.java
@@ -132,6 +132,22 @@ public class JournalParams {
           JournalRecord.ActionStatus.COMPLETED));
       }
     },
+    DI_INVENTORY_AUTHORITY_MATCHED {
+      @Override
+      public Optional<JournalParams> getJournalParams(DataImportEventPayload eventPayload) {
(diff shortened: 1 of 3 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/JournalParams.java
```diff
@@ -132,6 +132,22 @@ public JournalParams(JournalRecord.ActionType journalActionType,
           JournalRecord.ActionStatus.COMPLETED));
       }
     },
+    DI_INVENTORY_AUTHORITY_MATCHED {
```
psmagin, line 135: I don't think that we should log this event because it's not the end of the flow. It will be consumed by mod-inventory.
Instead of this event please add DI_INVENTORY_AUTHORITY_UPDATED event that is the end of the flow

psmagin (approved):

psmagin (approved):

viacheslavkol (approved):

PR #550, opened 2022-01-10
MODSOURMAN-664 Remove Kafka cache from QuickMarc handlers

Shans-Kaluhin:
## Purpose
Kafka Cache is used in QuickMarcUpdateKafkaHandler in SRM. Just remove Kafka cache (consumer receiving QM_COMPLETED supports deduplication).

## Learning
[MODSOURMAN-664](https://issues.folio.org/browse/MODSOURMAN-664)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/QuickMarcUpdateKafkaHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/QuickMarcUpdateKafkaHandler.java
@@ -1,10 +1,9 @@
 package org.folio.verticle.consumers;
 
-import static java.lang.String.format;
-
 import static org.folio.kafka.KafkaHeaderUtils.kafkaHeadersToMap;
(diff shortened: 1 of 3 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/QuickMarcUpdateKafkaHandler.java
```diff
@@ -44,32 +43,30 @@
   private static final String UNZIP_ERROR_MESSAGE = "Error during unzip";
   private static final String EVENT_ID_PREFIX = QuickMarcUpdateKafkaHandler.class.getSimpleName();
 
-  private final KafkaInternalCache kafkaInternalCache;
+  private final EventProcessedService eventProcessedService;
```
psmagin, line 46: There is no need for using this service. We can just remove the Kafka cache and rely on the database and handler that we have in mod-quick-marc.

psmagin (changes requested):

SerhiiNosko (approved):

psmagin (approved):

PR #552, opened 2022-01-12
MODSOURMAN-641: Remove Kafka cache by handling Constraint Violation Exceptions.

VRohach:
## Purpose
Kafka Cache is used in SRM in DataImportKafkaHandler and it should be replaced by another solution for handling duplicates bu DB.
## Approach
Used previously added table events_processed from story: MODSOURMAN-638
Populated events_processed table with value DataImportKafkaHandler after DataImportVerticle starts.
Modify DataImportKafkaHandler to catch ConstraintViolation exception and write into log about duplicate event.
## Learning
JIRA: https://issues.folio.org/browse/MODSOURMAN-641


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/DataImportKafkaHandler.java
```diff
@@ -27,17 +28,18 @@
...
 
   private Vertx vertx;
   private EventHandlingService eventHandlingService;
-  private KafkaInternalCache kafkaInternalCache;
+  private EventProcessedService eventProcessedService;
 
   public DataImportKafkaHandler(@Autowired Vertx vertx,
                                 @Autowired EventHandlingService eventHandlingService,
-                                @Autowired KafkaInternalCache kafkaInternalCache) {
+                                @Autowired @Qualifier("eventProcessedService") EventProcessedService eventProcessedService) {
```
SerhiiNosko, line 39: Qualifier annotation is not required here since only 1 implementation of EventProcessedService exists in context

VRohach: @SerhiiNosko Good point. Done.

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #555, opened 2022-01-13
MODSOURMAN-639 Improve performance of saving journal records during import

AivarIUsupov:
## Purpose
Need modification of method saveBatch() in JournalServiceImpl to improve performance.

## Approach
Instead of saving entities one by one in the cycle - we use execute() method from PostgresClient class (RMB), because currently for standard chunk size of 50 records we make 50 requests to DB instead of desired single call.

## Learning
[MODSOURMAN-639](https://issues.folio.org/browse/MODSOURMAN-639)


mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -129,6 +130,25 @@
       .onFailure(e -> LOGGER.error("Error saving JournalRecord entity", e));
   }
 
+  @Override
+  public Future<List<RowSet<Row>>> saveBatch(JsonArray entities, String tenantId) {
+    Promise<List<RowSet<Row>>> promise = Promise.promise();
+    try {
+      String query = format(INSERT_SQL, convertToPsqlStandard(tenantId), JOURNAL_RECORDS_TABLE);
+      List<Tuple> tupleList = new ArrayList<>();
+      for (int i = 0; i < entities.size(); i++) {
```
SerhiiNosko, line 139: As for me its better to make this conversion in the service layer, and DAO should accept just List<JornalRecord> journalRecords

AivarIUsupov: @SerhiiNosko Done, thanks!

Abdulkhakimov (approved):

RuslanLavrov (approved):

SerhiiNosko (approved):

PR #556, opened 2022-01-13
MODSOURMAN-640. Remove Kafka cache by handling Constraint Violation Exceptions for DataImportJournalKafkaHandler

SerhiiNosko:
## Purpose
Remove Kafka cache and at the same time need to support deduplication

## Approach
Replace Kafka cache with another deduplication implementation based on DB events_processed table

## Learning
https://issues.folio.org/browse/MODSOURMAN-640

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/DataImportJournalKafkaHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/DataImportJournalKafkaHandler.java
@@ -11,10 +11,11 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.DataImportEventPayload;
 import org.folio.dataimport.util.OkapiConnectionParams;
+import org.folio.dataimport.util.exception.ConflictException;
 import org.folio.kafka.AsyncRecordHandler;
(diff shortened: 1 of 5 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/DataImportJournalKafkaHandler.java
```diff
@@ -55,20 +55,32 @@ public DataImportJournalKafkaHandler(@Autowired Vertx vertx,
...
+      eventTypeHandlerSelector.getHandler(eventPayload).handle(journalService, eventPayload, tenantId);
       result.complete(record.key());
     } catch (Exception e) {
       LOGGER.error("Error during processing journal event", e);
       result.fail(e);
     }
-    return result.future();
   }
 
+  private void processDeduplicationFailure(Promise<String> result, KafkaConsumerRecord<String, String> record, Event event, Throwable e) {
```
SerhiiNosko, line 77: DataImportJournalKafkaHandler does not have error handler and there is no sense to create it, because DataImportJournalKafkaHandler  should not send DI_ERROR events, so in case of duplicates exception - completes future in this handler

Aliaksandr-Fedasiuk (approved):

KaterynaSenchenko (approved):

PR #565, opened 2022-01-27
MODSOURMAN-676 Provide Instance UUID for populating Inventory hotlinks for holdings/items

AivarIUsupov:
## Purpose
Extend functionality for /metadata-provider/jobLogEntries endpoint to provide information about Instance even if no action was performed on Instance.

## Approach
Added two new columns for journal_records table (instance_id and holdings_id) and implemented functionality to work with them.

## Learning
[MODSOURMAN-676](https://issues.folio.org/browse/MODSOURMAN-676)

## Information 
There are also known issue:
For the case, when we are going to only update ITEM entity with matching on ITEM's field, we will not get hotlink due to no information about INSTANCE entity (instanceId and holdingId are required to build hotlink for the ITEM entity).

### DIFF
--- a/mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
+++ b/mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
@@ -7,161 +7,168 @@ AS $$
 BEGIN
     RETURN QUERY
         SELECT records_actions.job_execution_id,
-               records_actions.source_id,
-               records_actions.source_record_order,
-               rec_titles.title,
-               CASE WHEN marc_errors_number != 0 OR marc_actions[array_length(marc_actions, 1)] = 'NON_MATCH'
-                        THEN 'DISCARDED'
-                    WHEN marc_actions[array_length(marc_actions, 1)] = 'CREATE'
-                        THEN 'CREATED'
-                    WHEN marc_actions[array_length(marc_actions, 1)] IN ('UPDATE', 'MODIFY')
-                        THEN 'UPDATED'
-               END AS source_record_action_status,
-			   records_actions.source_entity_error[1],
-               get_entity_status(instance_actions, instance_errors_number) AS instance_action_status,
-			   records_actions.instance_entity_id,
-			   records_actions.instance_entity_hrid,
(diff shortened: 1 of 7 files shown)


mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
```diff
@@ -7,161 +7,168 @@ AS $$
 BEGIN
```
SerhiiNosko, line 7: Its hard to see the difference with code formatting applied in this sql file, maybe revert back formatting?

AivarIUsupov: Thank  you @SerhiiNosko for the mention! To hide whitespaces you could use "Hide whitespaces" feature
Screen Shot 2022-01-28 at 11 59 42 AM
:

Aliaksandr-Fedasiuk (approved):

SerhiiNosko (approved):

Abdulkhakimov (approved):

VRohach (approved):

RuslanLavrov (approved):

mod-source-record-manager-server/src/test/java/org/folio/services/JournalUtilTest.java
```diff
@@ -13,14 +13,16 @@
 import java.util.HashMap;
 import java.util.UUID;
 
-import static org.folio.rest.jaxrs.model.JournalRecord.EntityType.INSTANCE;
-import static org.folio.rest.jaxrs.model.JournalRecord.EntityType.MARC_BIBLIOGRAPHIC;
+import static org.folio.rest.jaxrs.model.JournalRecord.ActionStatus.COMPLETED;
+import static org.folio.rest.jaxrs.model.JournalRecord.ActionType.CREATE;
+import static org.folio.rest.jaxrs.model.JournalRecord.ActionType.NON_MATCH;
+import static org.folio.rest.jaxrs.model.JournalRecord.EntityType.*;
```
RuslanLavrov, line 19: please don't use wildcard import

PR #569, opened 2022-02-04
MODINVOICE-356. Fix progress bar stuck behaviour after the RecordTooLargeException

SerhiiNosko:
## Purpose
Validate big chunks payload at the initial stage in SRM. If chunk payload size greater that Kafka MAX_REQUEST_SIZE - DI_ERROR event should be send. This DI_ERROR payload should contain enought data to correctly show log entries in View All and Details pages. 

## Approach
Implement specific DI_ERROR payload builders for each record type. Currently implemented for MARC-BIB and EDIFACT. Raw chunks error handler(initial stage of chunk processing in SRM) use these payload builders and send DI_ERROR events.
Each payload builder should take care to make DI_ERROR payload small as possible to prevent RecordTooLargeException when sending DI_ERROR events.
This logic applied not only for RecordTooLargeException  but also for other Kafka sending related exceptions.

#### Examples in UI
**When MARC_BIB import fails due to RecordTooLargeException  'View All' page would look like this:**
image
**When MARC_BIB import fails due to RecordTooLargeException  'Details page' would look like this:**
image

**When EDIFACT import fails due to RecordTooLargeException  'View All' page would look like this:**
image
**When EDIFACT import fails due to RecordTooLargeException  'Details page' would look like this:**
image

## Learning
https://issues.folio.org/browse/MODINVOICE-356)


Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/RecordsPublishingServiceImpl.java
```diff
@@ -12,7 +12,7 @@
 import org.folio.kafka.KafkaHeaderUtils;
 import org.folio.okapi.common.GenericCompositeFuture;
 import org.folio.rest.jaxrs.model.*;
```
KaterynaSenchenko, line 14: please remove this wildcard

KaterynaSenchenko (approved):

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/RecordsPublishingServiceImpl.java
SerhiiNosko: Done, thanks

PR #586, opened 2022-03-01
MODSOURMAN-714 Legacy 999 (non-ff) fields cause data import failure

AivarIUsupov:
## Purpose
Using data import to overlay a record with legacy 999 other than the one FOLIO creates with the inventory and srs UUIDs fields fails silently.

## Approach
Added additional validation for ff fields.

## Learning
[MODSOURMAN-714](https://issues.folio.org/browse/MODSOURMAN-714)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -76,6 +76,7 @@ import org.folio.services.parsers.RecordParserBuilder;
 
 import static org.folio.services.afterprocessing.AdditionalFieldsUtil.TAG_999;
 import static org.folio.services.afterprocessing.AdditionalFieldsUtil.addFieldToMarcRecord;
+import static org.folio.services.afterprocessing.AdditionalFieldsUtil.hasIndicator;
 import static org.folio.services.afterprocessing.AdditionalFieldsUtil.getControlFieldValue;
 import static org.folio.services.afterprocessing.AdditionalFieldsUtil.getValue;
 import static org.folio.services.util.EventHandlingUtil.sendEventToKafka;
@@ -271,9 +272,9 @@ public class ChangeEngineServiceImpl implements ChangeEngineService {
   }
 
   public List<Record> getParsedRecordsFromInitialRecords(List<InitialRecord> rawRecords,
-                                                          RecordsMetadata.ContentType recordContentType,
-                                                          JobExecution jobExecution,
-                                                          String sourceChunkId) {
+                                                         RecordsMetadata.ContentType recordContentType,
+                                                         JobExecution jobExecution,
+                                                         String sourceChunkId) {
     var parser = RecordParserBuilder.buildParser(recordContentType);
 
     return rawRecords.stream()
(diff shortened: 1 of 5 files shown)


SerhiiNosko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -426,8 +427,10 @@ private SourceStorageBatchClient getSourceStorageBatchClient(OkapiConnectionPara
   private void postProcessMarcRecord(Record record, InitialRecord rawRecord) {
     String matchedId = getValue(record, TAG_999, 's');
     if (StringUtils.isNotBlank(matchedId)) {
```
Abdulkhakimov, line 429: Can we merge the two if condition into one?

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -292,6 +292,17 @@ private static MarcReader buildMarcReader(Record record) {
     return new MarcJsonReader(new ByteArrayInputStream(record.getParsedRecord().getContent().toString().getBytes(StandardCharsets.UTF_8)));
   }
 
+  public static boolean hasIndicator(Record record, char subfield) {
+    MarcReader reader = buildMarcReader(record);
+    if (reader.hasNext()) {
+      org.marc4j.marc.Record marcRecord = reader.next();
+      VariableField variableField = getSingleFieldByIndicators(marcRecord.getVariableFields(TAG_999), INDICATOR, INDICATOR);
+      return variableField != null
```
Abdulkhakimov, line 300: I'm recomending use Objects.nonNull

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -426,8 +427,10 @@ private SourceStorageBatchClient getSourceStorageBatchClient(OkapiConnectionPara
...
-      record.setMatchedId(matchedId);
-      record.setGeneration(null); // in case the same record is re-imported, generation should be calculated on SRS side
+      if (hasIndicator(record, 's')) {
```
okolawole-ebsco, line 430: It would be good if these magic string like 's' or 'i' to be saved in a final variable and reused in other locations. It will also convey more meaning based on its more descriptive name.

AivarIUsupov: Totally agree, I've made changes for that.

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
AivarIUsupov: obj != null and Objects.nonNull are the same. I used this wrapper method to make it more clear, thank you @Abdulkhakimov !

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
AivarIUsupov: Hi @Abdulkhakimov. Done, thank you!

okolawole-ebsco (approved):

PR #594, opened 2022-03-17
MODSOURMAN-724 SRM does not process and save error records

Abdulkhakimov:
## Purpose
Link should be clickable when we click, but when SRS send DI error event records without parsed record have not information about title.

## Approach

- Added default title when we receive title without any data

## Learning
[MODSOURMAN-724](https://issues.folio.org/browse/MODSOURMAN-724)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
@@ -33,6 +33,8 @@ public class MarcImportEventsHandler implements SpecificEventHandler {
 
   public static final String INSTANCE_TITLE_FIELD_PATH = "title";
 
+  private static final String NO_MARC_TITLE_MESSAGE = "No content";
+
   private static final Map<JournalRecord.EntityType, BiFunction<ParsedRecord, JsonObject, String>> titleExtractorMap =
     Map.of(
       MARC_BIBLIOGRAPHIC, marcBibTitleExtractor(),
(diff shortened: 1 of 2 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
```diff
@@ -104,6 +106,8 @@ public void handle(JournalService journalService, DataImportEventPayload eventPa
   private Future<JournalRecord> populateRecordTitleIfNeeded(JournalRecord journalRecord,
                                                             DataImportEventPayload eventPayload) {
     var entityType = journalRecord.getEntityType();
+    Optional.ofNullable(journalRecord.getTitle())
+      .ifPresentOrElse(title -> {}, () -> journalRecord.setTitle(NO_MARC_TITLE_MESSAGE));
```
SerhiiNosko, line 110: I think we can simplify this, as option StringUtils and java trinary operator

SerhiiNosko, line 110: In the method we use titleExtractor to populate journalRecord with title, so I think better to add this check at the end of method

Abdulkhakimov: Done

Abdulkhakimov: Done thank you

SerhiiNosko (approved):

AivarIUsupov (approved):

RuslanLavrov (approved):

PR #605, opened 2022-03-29
MODSOURMAN-744: mapping_rules migration fails on missing "mappingRule"

julianladisch:
Add unique constraint on mapping_rules.unique_record_type to prevent duplicate records.

Provide a script to run manually that delete bogus rules: Rules without "mappingRules"
property in their jsonb field; and all but the first (smallest id) rule of that recordType
with "mappingRules" property.

## Learning
Add unique constraint from the beginning on. This avoids cleanup work that takes much more effort.


dcrossleyau:
That error from GitHub was transient and can be ignored. This repo is one of those still remaining to move to api-lint (but has various issues with JSON samples).

VRohach:
Hi @julianladisch! I think that we don't need this script in our codebase at all. As I already said, this script should be run manually for a specific environment and data, see comment: https://issues.folio.org/browse/MODSOURMAN-744

PR #620, opened 2022-04-15
[MODSOURMAN-662]Implement flow control logic in SRM for Kafka messages processing

SerhiiNosko:
## Purpose
Add flow control mechanism to be able to pause/resume processing of DI_RAW_RECORDS_CHUNK_READ topic. This will add possibility to process OCLC single record imports between main large imports.

## Approach
RawRecordsFlowControlService has been introduced, this will regulate load for DI_RAW_RECORDS_CHUNK_READ topic to process in batches(only after receiving DI_COMPLETE/DI_ERROR next batch reading from DI_RAW_RECORDS_CHUNK_READ). OCLC  processing skips topic DI_RAW_RECORDS_CHUNK_READ, flow for them starts from DI_PARSED_RECORDS_CHUNK_SAVED topic and due to flow control of DI_RAW_RECORDS_CHUNK_READ topic, OCLC imports processing in between of large data import jobs.
Currently Kafka consumers subscribed to all tenants, so pause/resume making also for all tenants that is a reason why couner if flow control is global(not per tenant).

Snippet for logs when importing 62 recs file with flow control enabled provided below(for this test chunk size equals to 10, 
max_simultaneous_records equals to 20, records_threshold equals to 10): 
image
image

## Learning
https://issues.folio.org/browse/MODSOURMAN-662


mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
```diff
@@ -0,0 +1,33 @@
+package org.folio.services.flowcontrol;
+
+/**
+ * Service to implement flow control logic to be able to import OCLC files in between imports of huge files.
+ * Its necessary to not wait for importing 1 record file until other big files in progress.
+ * Its can be suitable not only for OCLC, but also for other files that skipping DI_RAW_RECORDS_CHUNK_READ stage
```
okolawole-ebsco, line 6: typo: it instead of its

In general, i may be missing the point of this line. Here is my interpretation, "it is suitable for not only OCLC import, but also other import types that can skip DI_RAW_RECORDS_CHUNK_READ stage and pushing messages directly to DI_RAW_RECORDS_CHUNK_PARSED topic"

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public interface RawRecordsFlowControlService {
+
+  /**
+   * Tracks each DI_RAW_RECORDS_CHUNK_READ event, this method can also pause processing
+   * of DI_RAW_RECORDS_CHUNK_READ topic when flow control conditions met.
+   *
+   * @param initialRecordsCount records count in the batch
+   */
+  void trackChunkReceivedEvent(Integer initialRecordsCount);
```
okolawole-ebsco, line 17: would this method be tracking one event or many events? This method could be plural:`trackChunkReceivedEvents`

okolawole-ebsco: if this is acceptable, other methods in this interface should be plural as well.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
```diff
@@ -0,0 +1,33 @@
...
+ * Its can be suitable not only for OCLC, but also for other files that skipping DI_RAW_RECORDS_CHUNK_READ stage
+ * and directly pushing messages to DI_RAW_RECORDS_CHUNK_PARSED topic.
+ */
+public interface RawRecordsFlowControlService {
+
+  /**
+   * Tracks each DI_RAW_RECORDS_CHUNK_READ event, this method can also pause processing
+   * of DI_RAW_RECORDS_CHUNK_READ topic when flow control conditions met.
+   *
+   * @param initialRecordsCount records count in the batch
```
okolawole-ebsco, line 15: I am not able to discern what `initialRecordsCount` is pointing to OR how it will play with `trackchinkReceivedEvent`. The method doesnt accept a batch. Where is the batch coming from?

When the questions above are answered, it could help me to know what additional comments could be added here to make this interface self-documenting.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+import org.springframework.stereotype.Service;
+import javax.annotation.PostConstruct;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHUNK_READ;
+
+@Service
+public class RawRecordsFlowControlServiceImpl implements RawRecordsFlowControlService {
+  private static final Logger LOGGER = LogManager.getLogger();
```
okolawole-ebsco, line 18: Will this return the root logger or the logger for this class?

Typically I would see `LogManager.getLogger(RawRecordsFlowcontrolServiceImpl.class)` or something like that.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+  @Value("${di.flow.max_simultaneous_records:100}")
+  private Integer maxSimultaneousRecords;
+  @Value("${di.flow.records_threshold:50}")
+  private Integer recordsThreshold;
+  @Value("${di.flow.control.enable:true}")
```
okolawole-ebsco, line 24: nitpick: based on the scheme above i think this parameter should be `di.flow.enable`. I have not seen other parameters yet, so i may rescind this nitpick. This will hold especially if `enable` is the only option under `control`.

okolawole-ebsco: `di.flow_control` might get your idea across.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+  private boolean enableFlowControl;
+
+  @Autowired
+  private KafkaConsumersStorage consumersStorage;
+
+  private final AtomicInteger atomicCurrent = new AtomicInteger(0);
```
okolawole-ebsco, line 30: A small comment describing what this atomic integer is used for will be good. Especially since its name is very generic.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    if (current >= maxSimultaneousRecords) {
+      List<KafkaConsumerWrapper<String, String>> rawRecordsReadConsumers = consumersStorage.getConsumersByEvent(DI_RAW_RECORDS_CHUNK_READ.value());
+
+      rawRecordsReadConsumers.forEach(consumer -> {
+        if (consumer.demand() > 0) {
+          consumer.pause();
+
+          LOGGER.info("Kafka consumer - id: {}, subscription - {} is paused, because {} exceeded {} max simultaneous records",
+            consumer.getId(), DI_RAW_RECORDS_CHUNK_READ.value(), this.atomicCurrent, maxSimultaneousRecords);
```
okolawole-ebsco, line 55: I think `this.atomicCurrent` should be replaced with `current` declared above. The atomic integer might be set to zero by another thread before this log message is sent.

okolawole-ebsco: Overall, rather than passing the current count, you could just say that the max simultaneous records has been exceeded.

okolawole-ebsco: also if the atomic integer is left as is. `this.atomicCurrent.get()` would be better actually retrieve the value.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+      atomicCurrent.set(0);
+      return;
+    }
+
+    LOGGER.info("--------------- Current value after complete event: {} ---------------", current);
+
+    resumeIfThresholdAllows(current);
+  }
+
+  private void resumeIfThresholdAllows(int current) {
```
okolawole-ebsco, line 95: Since this method is only used in this class, it could work with the atomic integer declared as a field in this class rather than accepting an argument to work with.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+  @Override
+  public void trackRecordCompleteEvent() {
+    if (!enableFlowControl) {
+      return;
+    }
+
+    int current = atomicCurrent.decrementAndGet();
+
+    if (atomicCurrent.get() < 0) {
```
okolawole-ebsco, line 84: Since we are dealing with an atomic update here I think [getAndUpdate](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/atomic/AtomicInteger.html#getAndUpdate(java.util.function.IntUnaryOperator)) should be used to represent this code block somehow.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    resumeIfThresholdAllows(current);
+  }
+
+  private void resumeIfThresholdAllows(int current) {
+    if (current <= recordsThreshold) {
+      List<KafkaConsumerWrapper<String, String>> rawRecordsReadConsumers = consumersStorage.getConsumersByEvent(DI_RAW_RECORDS_CHUNK_READ.value());
+
+      rawRecordsReadConsumers.forEach(consumer -> {
+        if (consumer.demand() == 0) {
+          consumer.resume();
```
okolawole-ebsco, line 101: can this call fail? do we need to handle some exception here?

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+          LOGGER.info("Kafka consumer - id: {}, subscription - {} is resumed, because {} exceeded {} threshold",
+            consumer.getId(), DI_RAW_RECORDS_CHUNK_READ.value(), this.atomicCurrent, recordsThreshold);
```
okolawole-ebsco, line 104: `this.atomicCurrent` should be `this.atomicCurrent.get()` i think.

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
```diff
@@ -30,41 +31,45 @@ public abstract class AbstractConsumersVerticle extends AbstractVerticle {
...
         .globalLoadSensor(globalLoadSensor)
         .subscriptionDefinition(subscriptionDefinition)
         .processRecordErrorHandler(getErrorHandler())
-        .build());
-    });
-    List<Future<Void>> futures = new ArrayList<>();
-    consumerWrappersList.forEach(consumerWrapper ->
+        .backPressureGauge(getBackPressureGauge())
+        .build();
+      kafkaConsumersStorage.addConsumer(event, consumerWrapper);
```
okolawole-ebsco, line 60: Curious how consumers in other instances of SRM play with each other. Thinking maybe there might be some rogue instance of SRM still feeding large data import events instead of letting OCLC imports through.

mod-source-record-manager-server/src/main/java/org/folio/verticle/RawMarcChunkConsumersVerticle.java
```diff
@@ -35,4 +37,14 @@ public ProcessRecordErrorHandler<String, String> getErrorHandler() {
     return this.errorHandler;
   }
 
+  @Override
+  public BackPressureGauge<Integer, Integer, Integer> getBackPressureGauge() {
+    /*
+     * The simplest implementation to turn of handling of consumer's load by folio-kafka-wrapper.
```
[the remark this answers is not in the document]
okolawole-ebsco: I dont think I am understanding the entire comment after re-reading. I am not sure if the simple implementation is used instead of flow control.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
okolawole-ebsco: Reading the code a little bit more, I think if this method is named this way, it would accept the event as a parameter. In the implementation, the initial record count can be gotten.

or the name of these methods should be like `incrementChunkReceivedCount(int recordCount)`

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorage.java
```diff
@@ -0,0 +1,33 @@
...
+   */
+  void addConsumer(String eventName, KafkaConsumerWrapper<String, String> consumer);
+
+  /**
+   * Gets consumer by event name.
+   *
+   * @param eventName the event name
+   * @return consumer wrapper by event name
+   */
+  List<KafkaConsumerWrapper<String, String>> getConsumersByEvent(String eventName);
```
okolawole-ebsco, line 25: nitpick: these methods should return `Collection<KafkaconsumerWrapper<String,String>>` instead. This api is returning a List which means it is a collection that an item can be added or removed. Some people even return unmodifiable collection to reduce instance of bugs. Also returning an Array is cheaper than a List, especially since we dont need add/remove functionality of List.

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorageImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Component
+public class KafkaConsumersStorageImpl implements KafkaConsumersStorage {
+  private static final Logger LOGGER = LogManager.getLogger();
+
+  private final Map<String, List<KafkaConsumerWrapper<String, String>>> consumerWrappersMap = Collections.synchronizedMap(new HashMap<>());
```
okolawole-ebsco, line 20: nitpick: I have not used sychronizedMap before, typically I use [ConcurrentHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) if i need synchronization with a hashmap.

According to this [article](https://www.baeldung.com/java-synchronizedmap-vs-concurrenthashmap) ConcurrentHashMap doesn't allow nulls and supposed to be faster than synchronizedMap. I am fine if this is left as is.

mod-source-record-manager-server/src/test/java/org/folio/services/RawRecordsFlowControlServiceImplTest.java
```diff
@@ -0,0 +1,178 @@
...
+  private RawRecordsFlowControlServiceImpl service;
+
+  @Before
+  public void setUp() {
+    ReflectionTestUtils.setField(service, "enableFlowControl", true);
+  }
+
+  @Test
+  public void shouldSkipWhenFlowControlDisabled() {
+    ReflectionTestUtils.setField(service, "enableFlowControl", false);
```
okolawole-ebsco, line 41: nitpick: maybe have package private setters and getters for these fields that are modified by reflection. I think there might be some spring util to set these fields in a @Before method.

mod-source-record-manager-server/src/test/java/org/folio/services/RawRecordsFlowControlServiceImplTest.java
```diff
@@ -0,0 +1,178 @@
...
+    ReflectionTestUtils.setField(service, "maxSimultaneousRecords", 10);
+    ReflectionTestUtils.setField(service, "recordsThreshold", 5);
+
+    service.trackChunkReceivedEvent(10);
+    service.trackChunkDuplicateEvent(10);
+    service.trackRecordCompleteEvent();
+
+    verifyNoMoreInteractions(kafkaConsumersStorage);
```
okolawole-ebsco, line 49: Just for my understanding...

Please explain why any interaction would occur with the mock at this point. I think it is the `verifyNoMoreInteractions` that is confusing me. if it was `verifyNoInteractions` it would make more sense to me haha, but you dont control that.

 So with all the tracking done, no interaction should occur with `kafkaConsumerStorage` because flow control is disabled?

mod-source-record-manager-server/src/test/java/org/folio/services/RawRecordsFlowControlServiceImplTest.java
SerhiiNosko: I am not prefer to introduce setter methods just because we need them in tests, so used 
ReflectionTestUtils library from spring-test project and also we use it in other tests aw well.
Basically for each test we need different input params so I've  not used @Before for it.

mod-source-record-manager-server/src/test/java/org/folio/services/RawRecordsFlowControlServiceImplTest.java
```diff
@@ -0,0 +1,178 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHUNK_READ;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RawRecordsFlowControlServiceImplTest {
```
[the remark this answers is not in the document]
SerhiiNosko: Thanks.

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorageImpl.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Component
+public class KafkaConsumersStorageImpl implements KafkaConsumersStorage {
+  private static final Logger LOGGER = LogManager.getLogger();
```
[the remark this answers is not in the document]
SerhiiNosko: LogManager.getLogger - is our common approach to declare loggers. The internal implementation is:
`public static Logger getLogger() {
        return getLogger(StackLocatorUtil.getCallerClass(2));
    }`

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
SerhiiNosko: Updated java doc according your suggestion

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorage.java
SerhiiNosko: Modified return type to Collection instead of List

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorageImpl.java
SerhiiNosko: Im ok with using ConcurrentHashMap, made this change

mod-source-record-manager-server/src/test/java/org/folio/services/RawRecordsFlowControlServiceImplTest.java
SerhiiNosko: Thanks, changed to verifyNoInteractions and added more comments in code to clarify this test.
The answer to the second question - yes, no interaction should occur with kafkaConsumerStorage because flow control is disabled

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/consumerstorage/KafkaConsumersStorageImpl.java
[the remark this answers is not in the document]
SerhiiNosko: I checked in debug one more time what this implementation returns: `getLogger(StackLocatorUtil.getCallerClass(2));`
- its class specific logger

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: Thanks, changed code to read from field in this class  rather than accepting an argument to work with to improve readability

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: changed to use current variable.
By the way, toString() implementation of AtomicInteger does the same:
image

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: toString() implementation of AtomicInteger does basically the same:
image

mod-source-record-manager-server/src/main/java/org/folio/verticle/RawMarcChunkConsumersVerticle.java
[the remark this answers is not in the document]
SerhiiNosko: Here is this simplest implementation to handle load of kafka consumers inf folio-kafka-wrapper:
https://github.com/folio-org/folio-kafka-wrapper/blob/master/src/main/java/org/folio/kafka/KafkaConsumerWrapper.java#L82
All our consumers use it in order to avoid big pressure to database, oom issues etc. 
For RawMarcChunk consumers we use another flow control mechanism introduced in current PR, so we need to disable this standard one from folio-kafka-wrapper(because if dont need to expect for example to resume consumer in folio-kafka-wrapper if we paused in our flow control service)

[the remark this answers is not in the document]
SerhiiNosko: Link where pause() used in folio-kafka-wrapper: https://github.com/folio-org/folio-kafka-wrapper/blob/master/src/main/java/org/folio/kafka/KafkaConsumerWrapper.java#L197
Link where resume used in folio-kafka-wrapper: https://github.com/folio-org/folio-kafka-wrapper/blob/master/src/main/java/org/folio/kafka/KafkaConsumerWrapper.java#L251

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
okolawole-ebsco: great!

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
okolawole-ebsco: good to know!

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
SerhiiNosko: For now flow control intended for OCLC imports, but design should be extendable for future. For example, Spitfire team going to use our DI flow for importing quick marcs. This quick marcs imports should also be processed in between of large data imports. To use the same mechanism quick marc app should put messages in subsequent topic in flow after DI_RAW_RECORDS_CHUNK_READ. This tactics will allow to import quick marcs quickly without any modifications from our side.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: renamed all 3 params according params convention from mod-data-import: [Link](https://github.com/folio-org/mod-data-import/blob/1b0aa10019ed45a463e143e461104944b27e8687/src/main/java/org/folio/service/processing/reader/SourceReaderBuilder.java#L18)
Now they looks like this:
di.flow.control.max.simultaneous.records
di.flow.control.records.threshold
di.flow.control.enable

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: Renamed this field to 'currentState' and added descriptive java doc for it.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private void resumeIfThresholdAllows(int current) {
+    if (current <= recordsThreshold) {
+      List<KafkaConsumerWrapper<String, String>> rawRecordsReadConsumers = consumersStorage.getConsumersByEvent(DI_RAW_RECORDS_CHUNK_READ.value());
+
+      rawRecordsReadConsumers.forEach(consumer -> {
+        if (consumer.demand() == 0) {
+          consumer.resume();
+
+          LOGGER.info("Kafka consumer - id: {}, subscription - {} is resumed, because {} exceeded {} threshold",
```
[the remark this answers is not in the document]
SerhiiNosko: A bit paraphrased this log, changed word 'exceeded' to 'met'

mod-source-record-manager-server/src/main/java/org/folio/verticle/RawMarcChunkConsumersVerticle.java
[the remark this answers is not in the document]
okolawole-ebsco: thanks for the explanation and references! it makes sense now.
If I may re-write the comment

> Disable back pressure gauge defined by folio-kafka-wrapper by setting this simple implementation. This implementation will not allow folio-kafka-wrapper to pause/resume the topic. Flow control mechanism, defined in this codebase, will be used instead to handle load from DI_RAW_RECORDS_CHUNK_READ topic. Flow control will have exclusive rights to pause/resume the topic.
@see RawRecordsFlowControlService

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
okolawole-ebsco: let me ask a different way.

Two instances of SRM are running. Both have a consumer defined on a topic to read from it. Would there be situations where both instances of SRM will try to pause the topic at the same time?

mod-source-record-manager-server/src/main/java/org/folio/verticle/RawMarcChunkConsumersVerticle.java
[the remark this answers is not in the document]
SerhiiNosko: Committed change, thanks

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
okolawole-ebsco: Awesome comment that describes things!

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
okolawole-ebsco: looks good!

okolawole-ebsco:
@SerhiiNosko The process implemented by this PR expects that each record in a MARC file will have a corresponding event in DI_RAW_RECORDS_CHUNK_READ and a DI_COMPLETE/DI_ERROR when all computation is done by our distributed data import system. What happens if a DI_COMPLETE/DI_ERROR is never received for a record in a file?

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Represents the current state of flow control to make decisions when need to make pause/resume calls.
+   * It is basically counter, that increasing when DI_RAW_RECORDS_CHUNK_READ event comes and decreasing when
+   * DI_COMPLETE, DI_ERROR come.
+   *
+   * When current state equals or exceeds di.flow_control.max_simultaneous_records - all raw records consumers from consumer's group should pause.
+   * When current state equals or below di.flow_control.records_threshold - all raw records consumers from consumer's group should resume.
+   */
+  private final AtomicInteger currentState = new AtomicInteger(0);
```
okolawole-ebsco, line 38: General question. is handling of message done on the vertx event loop?

trying to reason to see if having this synchronization will slow things.

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
SerhiiNosko: For example, we have 2 instances of SRM. Each instance configured to have 3 consumers in consumer group to read from DI_RAW_RECORDS_CHUNK_READ topic by default.
Each instance would have separate counter to track internal state of flow control. If for example max.simultaneous.records met for instance №1 - it will be paused, but 3 consumers from instance №2 would continue reading.
We pausing only consumers, producers always working.

SerhiiNosko:
> @SerhiiNosko The process implemented by this PR expects that each record in a MARC file will have a corresponding event in DI_RAW_RECORDS_CHUNK_READ and a DI_COMPLETE/DI_ERROR when all computation is done by our distributed data import system. What happens if a DI_COMPLETE/DI_ERROR is never received for a record in a file?

SerhiiNosko:
> > @SerhiiNosko The process implemented by this PR expects that each record in a MARC file will have a corresponding event in DI_RAW_RECORDS_CHUNK_READ and a DI_COMPLETE/DI_ERROR when all computation is done by our distributed data import system. What happens if a DI_COMPLETE/DI_ERROR is never received for a record in a file?

Thanks for this question, I also always thinking about it and concerned about it. There can be situation when import stucks, DI_COMPLETE/DI_ERROR never sent and consumers remain always in paused state. To handle this I am thinking to add @Scheduled method that  will reset internal state and resume consumer if needed each minute.
Because our main goal - avoid situation when consumer paused forever and control the load in the same time.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
SerhiiNosko: we receiving messages from DI_RAW_RECORDS_CHUNK_READ topic in chunks, it is one event that contain chunk with multiple records. For DI_COMPLETE/DI_ERROR we always operate with individual records(not chunks).
Based on this we have words 'chunk' and 'record' in method names.
I have not added something like 'count' in method name to not associate it with simple counter. 
I have not added event as parameter to simplify things, if we not use this object in method - we dont need to construct full event and pass as param.
@okolawole-ebsco  what do you think, can we remain method names as it or you have additional suggestions?

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    int current = atomicCurrent.addAndGet(initialRecordsCount);
+
+    LOGGER.info("--------------- Current value after chunk received: {} ---------------", current);
+
+    if (current >= maxSimultaneousRecords) {
+      List<KafkaConsumerWrapper<String, String>> rawRecordsReadConsumers = consumersStorage.getConsumersByEvent(DI_RAW_RECORDS_CHUNK_READ.value());
+
+      rawRecordsReadConsumers.forEach(consumer -> {
+        if (consumer.demand() > 0) {
+          consumer.pause();
```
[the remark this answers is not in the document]
SerhiiNosko: We don't need to handle exceptions here, I've added @Scheduled method that invoked each 5 min by default that will reset flow control state and back consumer to resume if it paused. It will help to avoid situation when consumer paused forever. More details you could see in java docs of resetState() method.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: We don't need to handle exceptions here, I've added @Scheduled method that invoked each 5 min by default that will reset flow control state and back consumer to resume if it paused. It will help to avoid situation when consumer paused forever. More details you could see in java docs of resetState() method.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: Thanks, rewritten to use getAndUpdate method

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
okolawole-ebsco: ok so the consumers are being paused, not the topic. I knew this but i was not using it in my thought process. Thanks for the clarification!

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
okolawole-ebsco: Thanks!

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
okolawole-ebsco: I dont have any other suggestions. Your explanation helped and the changes you made seems more correct to me than the previous version.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlService.java
okolawole-ebsco: The latest refactor of this interface has cleared things for me.

VRohach (approved):
Awesome work @SerhiiNosko !

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @Override
+  public void trackChunkDuplicateEvent(Integer recordsCount) {
+    if (!enableFlowControl) {
+      return;
+    }
+
+    int prev = currentState.get();
+
+    currentState.set(prev - recordsCount);
```
okolawole-ebsco, line 107: i think getAndUpdate should be used to perform the previous two lines atomically.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -0,0 +1,149 @@
...
+
+    LOGGER.info("--------------- Chunk duplicate event comes, update current value from: {} to: {} ---------------", prev, currentState.get());
```
okolawole-ebsco, line 109: This log would be in the getAndUpdate i suggested above.

SerhiiNosko: Moved log to getAndUpdate() function

PR #621, opened 2022-04-20
MODSOURMAN-771 - New API information for DI Delete Logs

SatishkumarPurswani:
## Purpose
This is a new API being introduced to delete single or multiple data-import logs

## Approach
Calling this API will allow deleting one or more data-import logs depending upon the information sent.

#### TODOS and Open Questions
- [x] Unable to test this API from my local, I will check this textbox once tested.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -122,6 +122,19 @@
         },
         {
           "methods": [
+            "DELETE"
+          ],
+          "pathPattern": "/change-manager/jobExecutions",
+          "permissionsRequired": [
+            "change-manager.jobexecutions.post"
+          ],
+          "modulePermissions": [
+            "source-storage.snapshots.post",
+            "users.collection.get"
+          ]
+        },
+        {
+          "methods": [
             "PUT"
           ],
           "pathPattern": "/change-manager/jobExecutions/{id}",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -126,11 +126,7 @@
           ],
           "pathPattern": "/change-manager/jobExecutions",
           "permissionsRequired": [
-            "change-manager.jobexecutions.post"
-          ],
-          "modulePermissions": [
-            "source-storage.snapshots.post",
-            "users.collection.get"
+            "change-manager.jobexecutions.delete"
           ]
         },
         {
@@ -484,6 +480,11 @@
       "permissionName": "change-manager.jobexecutions.post",
       "displayName": "Change Manager - create jobExecutions",
       "description": "Post JobExecution"
+    },
+    {
+      "permissionName": "change-manager.jobexecutions.delete",
+      "displayName": "Change Manager - delete jobExecutions",
+      "description": "Delete JobExecutions Using JobExecutionIds"
     },
     {
       "permissionName": "change-manager.jobexecutions.put",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -484,7 +484,7 @@
     {
       "permissionName": "change-manager.jobexecutions.delete",
       "displayName": "Change Manager - delete jobExecutions",
-      "description": "Delete JobExecutions Using JobExecutionIds"
+      "description": "Batch Delete JobExecutions"
     },
     {
       "permissionName": "change-manager.jobexecutions.put",
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
@@ -71,5 +71,11 @@ public interface JobExecutionDao {
    */
   Future<JobExecution> updateBlocking(String jobExecutionId, JobExecutionMutator mutator, String tenantId);
 
-  Future<Boolean> deleteJobExecutionByIds(List<String> ids, String tenantId);
+  /**
+   *
+   * @param ids JobExecutionIds to be deleted ids
+   * @param tenantId
+   * @return future of boolean depending upon success and failure
+   */
+  Future<Boolean> deleteJobExecutionsByIds(List<String> ids, String tenantId);
 }
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
@@ -3,7 +3,7 @@ package org.folio.dao;
 import io.vertx.core.Future;
 import org.folio.dao.util.JobExecutionMutator;
 import org.folio.dao.util.SortField;
-import org.folio.rest.jaxrs.model.DeleteJobExecutionsRespDto;
+import org.folio.rest.jaxrs.model.DeleteJobExecutionsResp;
 import org.folio.rest.jaxrs.model.JobExecution;
 import org.folio.rest.jaxrs.model.JobExecutionDtoCollection;
 
@@ -78,5 +78,5 @@ public interface JobExecutionDao {
    * @param tenantId
    * @return future of boolean depending upon success and failure
    */
-  Future<DeleteJobExecutionsRespDto> softDeleteJobExecutionsByIds(List<String> ids, String tenantId);
+  Future<DeleteJobExecutionsResp> softDeleteJobExecutionsByIds(List<String> ids, String tenantId);
 }
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
@@ -71,4 +71,5 @@ public interface JobExecutionDao {
    */
   Future<JobExecution> updateBlocking(String jobExecutionId, JobExecutionMutator mutator, String tenantId);
 
+  Future<Boolean> deleteJobExecutionByIds(List<String> ids, String tenantId);
 }
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
@@ -3,6 +3,7 @@ package org.folio.dao;
 import io.vertx.core.Future;
 import org.folio.dao.util.JobExecutionMutator;
 import org.folio.dao.util.SortField;
+import org.folio.rest.jaxrs.model.DeleteJobExecutionsRespDto;
 import org.folio.rest.jaxrs.model.JobExecution;
 import org.folio.rest.jaxrs.model.JobExecutionDtoCollection;
 
@@ -77,5 +78,5 @@ public interface JobExecutionDao {
    * @param tenantId
    * @return future of boolean depending upon success and failure
    */
-  Future<Boolean> deleteJobExecutionsByIds(List<String> ids, String tenantId);
+  Future<DeleteJobExecutionsRespDto> softDeleteJobExecutionsByIds(List<String> ids, String tenantId);
 }
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
@@ -73,7 +73,7 @@ public interface JobExecutionDao {
 
   /**
    *
-   * @param ids JobExecutionIds to be deleted ids
+   * @param ids JobExecution Logs to be deleted using Ids
    * @param tenantId
    * @return future of boolean depending upon success and failure
    */
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
@@ -37,6 +37,7 @@ import static java.util.Objects.nonNull;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
 import static org.folio.dao.util.JobExecutionDBConstants.COMPLETED_DATE_FIELD;
 import static org.folio.dao.util.JobExecutionDBConstants.CURRENTLY_PROCESSED_FIELD;
+import static org.folio.dao.util.JobExecutionDBConstants.DELETE_BY_IDS_SQL;
 import static org.folio.dao.util.JobExecutionDBConstants.ERROR_STATUS_FIELD;
 import static org.folio.dao.util.JobExecutionDBConstants.FILE_NAME_FIELD;
 import static org.folio.dao.util.JobExecutionDBConstants.GET_BY_ID_SQL;
@@ -203,6 +204,20 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
         promise.complete(jobExecutionPromise.future().result());
       });
     return promise.future();
+  }
+
+  @Override
+  public Future<Boolean> deleteJobExecutionByIds(List<String> ids, String tenantId) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    try {
+      String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME);
+      Tuple queryParams = Tuple.from(ids);
+      pgClientFactory.createInstance(tenantId).delete(query, queryParams, promise);
+    } catch (Exception e) {
+      LOGGER.error("Error getting jobExecution by id", e);
+      promise.fail(e);
+    }
+    return promise.future().map(updateResult -> updateResult.rowCount() == 1);
   }
 
   private Tuple mapToTuple(JobExecution jobExecution) {
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
@@ -207,7 +207,7 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
   }
 
   @Override
-  public Future<Boolean> deleteJobExecutionByIds(List<String> ids, String tenantId) {
+  public Future<Boolean> deleteJobExecutionsByIds(List<String> ids, String tenantId) {
     Promise<RowSet<Row>> promise = Promise.promise();
     try {
       String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME, ids.stream().collect(Collectors.joining("','")));
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
@@ -213,10 +213,10 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
       String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME, ids.stream().collect(Collectors.joining("','")));
       pgClientFactory.createInstance(tenantId).execute(query, promise);
     } catch (Exception e) {
-      LOGGER.error("Error deleting jobExecution by ids {} : Exception : ", ids, e);
+      LOGGER.error("Error deleting jobExecution by ids {}, ", ids, e);
       promise.fail(e);
     }
-    return promise.future().map(updateResult -> updateResult.rowCount() == 1);
+    return promise.future().map(updateResult -> updateResult.rowCount()>0);
   }
 
   private Tuple mapToTuple(JobExecution jobExecution) {
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
@@ -210,9 +210,8 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
   public Future<Boolean> deleteJobExecutionByIds(List<String> ids, String tenantId) {
     Promise<RowSet<Row>> promise = Promise.promise();
     try {
-      String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME);
-      Tuple queryParams = Tuple.from(ids);
-      pgClientFactory.createInstance(tenantId).delete(query, queryParams, promise);
+      String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME, ids.stream().collect(Collectors.joining("','")));
+      pgClientFactory.createInstance(tenantId).execute(query, promise);
     } catch (Exception e) {
       LOGGER.error("Error getting jobExecution by id", e);
       promise.fail(e);
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
@@ -88,6 +88,8 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
   private static final String TABLE_NAME = "job_execution";
   private static final String PROGRESS_TABLE_NAME = "job_execution_progress";
   public static final String GET_JOB_EXECUTION_HR_ID = "SELECT nextval('%s.job_execution_hr_id_sequence')";
+
+  //Below constants are used for building db query related to job execution deletions
   public static final String ID = "id";
   public static final String IS_DELETED = "is_deleted";
   public static final String RETURNING_FIELD_NAMES = "returningFieldNames";
@@ -226,7 +228,7 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
   public Future<DeleteJobExecutionsResp> softDeleteJobExecutionsByIds(List<String> ids, String tenantId) {
     Promise<RowSet<Row>> promise = Promise.promise();
     try {
-      Map data = new HashMap<String, String>();
+      Map<String, String> data = new HashMap<>();
       data.put(TENANT_NAME, convertToPsqlStandard(tenantId));
       data.put(DB_TABLE_NAME_FIELD,TABLE_NAME);
       data.put(SET_FIELD_NAME, IS_DELETED);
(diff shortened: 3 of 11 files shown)


mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
```diff
@@ -71,4 +71,5 @@ public interface JobExecutionDao {
    */
   Future<JobExecution> updateBlocking(String jobExecutionId, JobExecutionMutator mutator, String tenantId);
 
+  Future<Boolean> deleteJobExecutionByIds(List<String> ids, String tenantId);
```
SerhiiNosko, line 74: You are deleting multiple job executions here so better to rename method such as:  deleteJobExecutionsByIds

SerhiiNosko: also better to add java docs such as other methods

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionService.java
```diff
@@ -114,4 +114,7 @@ public interface JobExecutionService {
    */
   Future<Boolean> completeJobExecutionWithError(String jobExecutionId, OkapiConnectionParams params);
 
+
+  Future<Boolean> deleteJobExecutionByIds(List<String> ids, String tenantId);
```
SerhiiNosko, line 118: The same with method renaming

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/changeManager/ChangeManagerAPITest.java
```diff
@@ -1885,4 +1886,48 @@ public void shouldSetErrorToRecordWithInvalidLeaderLine(TestContext testContext)
...
+    RestAssured.given()
+      .spec(spec)
+      .body("{\"ids\":[\"15065ccd-c305-4961-a411-8359e0b5a87b\", \"15065ccd-c305-4961-a411-8359e0b5a87b\"]}")
+      .when()
+      .delete(JOB_EXECUTION_PATH)
+      .then()
+      .statusCode(HttpStatus.SC_OK);
+  }
+
+  @Test @Ignore
```
SerhiiNosko, line 1911: don't forget to delete @Ignore annotation

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -51,6 +52,24 @@ public ChangeManagerImpl(Vertx vertx, String tenantId) { //NOSONAR
...
+
+  @Override
+  public void deleteChangeManagerJobExecutions(DeleteJobExecutionsDto entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    vertxContext.runOnContext(c -> {
+      try {
+        jobExecutionService.deleteJobExecutionByIds(entity.getIds(), tenantId)
+          .map(optionalJobExecution -> !optionalJobExecution.booleanValue()?
+            new BadRequestException(format("JobExecution with id '%s' was not found", entity.getIds())): optionalJobExecution)
+          .map(Response.class::cast)
+          .otherwise(Response.ok().build())
```
SerhiiNosko, line 64: Is it possible to rewrite, because otherwise() more suitable for failure cases, not success ones

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/changeManager/ChangeManagerAPITest.java
SatishkumarPurswani: @SerhiiNosko - PR updated as per comments, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
SatishkumarPurswani: @SerhiiNosko - PR updated as per comments, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionService.java
SatishkumarPurswani: @SerhiiNosko - PR updated as per comments, Thanks 🙂

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/changeManager/ChangeManagerAPITest.java
```diff
@@ -78,6 +78,7 @@
 import org.hamcrest.MatcherAssert;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
```
SerhiiNosko, line 81: This unnecessary import can be deleted

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -51,6 +52,25 @@ public ChangeManagerImpl(Vertx vertx, String tenantId) { //NOSONAR
...
+      try {
+        jobExecutionService.deleteJobExecutionsByIds(entity.getIds(), tenantId)
+          .map(optionalJobExecution -> !optionalJobExecution.booleanValue()?
+            new BadRequestException(format("JobExecution with id '%s' was not found", entity.getIds())): optionalJobExecution)
+          .map(DeleteChangeManagerJobExecutionsResponse::respond200WithApplicationJson)
+          .map(Response.class::cast)
+          .otherwise(ExceptionHelper::mapExceptionToResponse)
+          .onComplete(asyncResultHandler);
+      } catch (Exception e) {
+        LOGGER.error(getMessage("Failed to get JobExecution by id", e, entity.getIds().get(0)));
```
SerhiiNosko, line 68: Can rewrite to more simple:
LOGGER.error("Failed to delete Job Executions by ids: {}",  entity.getIds(), e);

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -205,6 +206,19 @@ public Future<JobExecution> updateBlocking(String jobExecutionId, JobExecutionMu
...
   }
 
+  @Override
+  public Future<Boolean> deleteJobExecutionsByIds(List<String> ids, String tenantId) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    try {
+      String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME, ids.stream().collect(Collectors.joining("','")));
+      pgClientFactory.createInstance(tenantId).execute(query, promise);
+    } catch (Exception e) {
+      LOGGER.error("Error getting jobExecution by id", e);
```
SerhiiNosko, line 216: Please correct message to: LOGGER.error("Error deleting job executions by ids: {}", ids, e);

descriptors/ModuleDescriptor-template.json
```diff
@@ -120,6 +120,19 @@
...
+          ],
+          "pathPattern": "/change-manager/jobExecutions",
+          "permissionsRequired": [
+            "change-manager.jobexecutions.post"
+          ],
+          "modulePermissions": [
+            "source-storage.snapshots.post",
+            "users.collection.get"
+          ]
+        },
```
SerhiiNosko, line 131: Permission set required to make deletion described somewhere like story AC?

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
RuslanLavrov, line 216: please change the message according to the operation

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/changeManager/ChangeManagerAPITest.java
SatishkumarPurswani: @SerhiiNosko - PR updated as per comments, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
SatishkumarPurswani: @SerhiiNosko - PR updated as per comments, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: @SerhiiNosko - PR updated as per comments, Thanks 🙂

descriptors/ModuleDescriptor-template.json
SatishkumarPurswani: @SerhiiNosko - per our discussion, will confirm with the required stakeholders and make necessary changes if required. Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -51,6 +52,25 @@ public ChangeManagerImpl(Vertx vertx, String tenantId) { //NOSONAR
...
+      try {
+        jobExecutionService.deleteJobExecutionsByIds(entity.getIds(), tenantId)
+          .map(optionalJobExecution -> !optionalJobExecution.booleanValue()?
+            new BadRequestException(format("JobExecution with id '%s' was not found", entity.getIds())): optionalJobExecution)
+          .map(DeleteChangeManagerJobExecutionsResponse::respond200WithApplicationJson)
+          .map(Response.class::cast)
+          .otherwise(ExceptionHelper::mapExceptionToResponse)
+          .onComplete(asyncResultHandler);
+      } catch (Exception e) {
+        LOGGER.error("Failed to get JobExecution by id {} : Exception : ", entity.getIds(), e);
```
SerhiiNosko, line 68:
```suggestion
        LOGGER.error("Failed to delete  JobExecutions  by ids: {}", entity.getIds(), e);
```

SatishkumarPurswani: @SerhiiNosko - PR updated as per comments, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -205,6 +206,19 @@ public Future<JobExecution> updateBlocking(String jobExecutionId, JobExecutionMu
...
   }
 
+  @Override
+  public Future<Boolean> deleteJobExecutionsByIds(List<String> ids, String tenantId) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    try {
+      String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME, ids.stream().collect(Collectors.joining("','")));
+      pgClientFactory.createInstance(tenantId).execute(query, promise);
+    } catch (Exception e) {
+      LOGGER.error("Error deleting jobExecution by ids {} : Exception : ", ids, e);
```
AivarIUsupov, line 216: Hi @SerhiiNosko @SatishkumarPurswani , we don't have "entity" here, so we can't use it.

mod-source-record-manager-server/src/main/java/org/folio/dao/util/JobExecutionDBConstants.java
```diff
@@ -28,6 +28,7 @@ public final class JobExecutionDBConstants {
   public static final String TOTAL_FIELD = "total";
 
   public static final String GET_BY_ID_SQL = "SELECT * FROM %s WHERE id = $1";
+  public static final String DELETE_BY_IDS_SQL = "DELETE FROM %s.%s WHERE id IN ( '%s' )";
```
AivarIUsupov, line 31: Hi @SatishkumarPurswani Is there any possibility to use Tuples instead of formatting String?

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionService.java
```diff
@@ -114,4 +114,13 @@ public interface JobExecutionService {
    */
   Future<Boolean> completeJobExecutionWithError(String jobExecutionId, OkapiConnectionParams params);
 
+
+  /**
+   *
+   * @param ids JobExecutionIds to be deleted ids
```
AivarIUsupov, line 120:
```suggestion
   * @param ids JobExecution ids to be deleted
```

SerhiiNosko (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -205,6 +206,19 @@ public Future<JobExecution> updateBlocking(String jobExecutionId, JobExecutionMu
     return promise.future();
   }
 
+  @Override
+  public Future<Boolean> deleteJobExecutionsByIds(List<String> ids, String tenantId) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    try {
+      String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME, ids.stream().collect(Collectors.joining("','")));
```
AivarIUsupov, line 213: Hi @SatishkumarPurswani Do we also need to clear job_execution_progress and job_execution_source_chunks tables to not keep "corrupted" data.

AivarIUsupov (changes requested):

mod-source-record-manager-server/src/main/java/org/folio/dao/util/JobExecutionDBConstants.java
SatishkumarPurswani: Hi @AivarIUsupov - I tried using tuple but it was not working with IN query, instead it was iterating over multiple tuples which means the number of queries will be equal to the number of JobExecution Ids that we have requested to delete.
Hence, had to use this approach.
Feel free to suggest a better approach, if any.

descriptors/ModuleDescriptor-template.json
```diff
@@ -472,6 +481,11 @@
       "displayName": "Change Manager - create jobExecutions",
       "description": "Post JobExecution"
     },
+    {
+      "permissionName": "change-manager.jobexecutions.delete",
+      "displayName": "Change Manager - delete jobExecutions",
+      "description": "Delete JobExecutions Using JobExecutionIds"
```
okolawole-ebsco, line 487: nit: Thinking this could be just "Delete jobExecutions". If you want to demonstrate that multiple ids can be accepted, you could include the word "batch" so maybe "Batch Delete JobExecutions"

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
okolawole-ebsco, line 216: It would be cleaner to remove "Exception : " from the log message string. You are not using the substitution token `{}`. The logger implementation we use appends the error message at the end, but it isn't always the case.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -205,6 +206,19 @@ public Future<JobExecution> updateBlocking(String jobExecutionId, JobExecutionMu
...
+      pgClientFactory.createInstance(tenantId).execute(query, promise);
+    } catch (Exception e) {
+      LOGGER.error("Error deleting jobExecution by ids {} : Exception : ", ids, e);
+      promise.fail(e);
+    }
+    return promise.future().map(updateResult -> updateResult.rowCount() == 1);
```
okolawole-ebsco, line 219: if more than one job execution is deleted would the rowCount reflect that? 
if so the condition should be `updateResult.rowCount() > 0`.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -51,6 +52,25 @@ public ChangeManagerImpl(Vertx vertx, String tenantId) { //NOSONAR
...
+      try {
+        jobExecutionService.deleteJobExecutionsByIds(entity.getIds(), tenantId)
+          .map(optionalJobExecution -> !optionalJobExecution.booleanValue()?
+            new BadRequestException(format("JobExecution with id '%s' was not found", entity.getIds())): optionalJobExecution)
+          .map(DeleteChangeManagerJobExecutionsResponse::respond200WithApplicationJson)
+          .map(Response.class::cast)
+          .otherwise(ExceptionHelper::mapExceptionToResponse)
+          .onComplete(asyncResultHandler);
+      } catch (Exception e) {
+        LOGGER.error("Failed to delete JobExecutions by ids {} : Exception : ", entity.getIds(), e);
```
okolawole-ebsco, line 68: "Exception :" is not needed here. It is not obvious from the api of LOGGER.error that the error message is appended at the end.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -51,6 +52,25 @@ public ChangeManagerImpl(Vertx vertx, String tenantId) { //NOSONAR
...
+        asyncResultHandler.handle(Future.succeededFuture(ExceptionHelper.mapExceptionToResponse(e)));
```
okolawole-ebsco, line 69: nit: should this be a Future.failedFuture instead?

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
okolawole-ebsco: I think have the log level as ERROR makes it easier. and also you could wrap all the ids in brackets like so `[ id1, id2 ]`. I dont think it would be too difficult to know where logging starts. Yes, please make the change.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
SatishkumarPurswani: Hi @okolawole-ebsco - Updated changes as per review comment, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionService.java
SatishkumarPurswani: Hi @AivarIUsupov - Updated changes as per review comment, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: Hi @okolawole-ebsco - Updated changes as per review comment, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: Hi @okolawole-ebsco - Updated changes as per review comment, Thanks 🙂

descriptors/ModuleDescriptor-template.json
SatishkumarPurswani: Hi @okolawole-ebsco - Updated changes as per review comment, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -213,10 +213,10 @@ public Future<Boolean> deleteJobExecutionsByIds(List<String> ids, String tenantI
       String query = format(DELETE_BY_IDS_SQL, convertToPsqlStandard(tenantId), TABLE_NAME, ids.stream().collect(Collectors.joining("','")));
       pgClientFactory.createInstance(tenantId).execute(query, promise);
     } catch (Exception e) {
-      LOGGER.error("Error deleting jobExecution by ids {} : Exception : ", ids, e);
+      LOGGER.error("Error deleting jobExecution by ids {}, ", ids, e);
       promise.fail(e);
     }
-    return promise.future().map(updateResult -> updateResult.rowCount() == 1);
+    return promise.future().map(updateResult -> updateResult.rowCount()>0);
```
okolawole-ebsco, line 219: nitpick: could you space this adequately? A spaces surrounding `>`

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
okolawole-ebsco: @SatishkumarPurswani what do you think of this?

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -51,6 +52,25 @@ public ChangeManagerImpl(Vertx vertx, String tenantId) { //NOSONAR
...
   }
 
+
+  @Override
+  public void deleteChangeManagerJobExecutions(DeleteJobExecutionsDto entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    vertxContext.runOnContext(c -> {
+      try {
+        jobExecutionService.deleteJobExecutionsByIds(entity.getIds(), tenantId)
+          .map(optionalJobExecution -> !optionalJobExecution.booleanValue()?
+            new BadRequestException(format("JobExecution with id '%s' was not found", entity.getIds())): optionalJobExecution)
```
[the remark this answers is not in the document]
SatishkumarPurswani: Hi @okolawole-ebsco - This works fine as 1st parameter is a type of list.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
SatishkumarPurswani: Hi @okolawole-ebsco - I think, we are returning the `Future.succeededFuture` as we want to pass the status code and the message along with it, whereas `Future.failedFuture` takes either String or Throwable which will not suffice the need.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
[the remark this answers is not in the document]
okolawole-ebsco: Sorry I was referring to the error message itself. Something like `JobExecutions with ids '%s' were not found`

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
okolawole-ebsco: Great point! Yeah, it is good as it is.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
SatishkumarPurswani: @SerhiiNosko - I have updated the PR as suggested.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: @AivarIUsupov - Changes have been done as per our new approach now, Thanks 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: @okolawole-ebsco - New code pushed in, above changes not required, Thanks 🙂.

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -81,6 +88,16 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
   private static final String TABLE_NAME = "job_execution";
   private static final String PROGRESS_TABLE_NAME = "job_execution_progress";
   public static final String GET_JOB_EXECUTION_HR_ID = "SELECT nextval('%s.job_execution_hr_id_sequence')";
+  public static final String ID = "id";
```
okolawole-ebsco, line 91: nitpick: i think it would be nice to add a comment that says that these constants are meant to be used for substitution into a sql query that is somewhere.

Ideally, I would want something like a query builder object that will accept these values. The query builder would substitute the value and return a complete valid SQL string. That way everything is modularized and self-contained. The current approach breaks if a substitution name is changed in the other DB constants class. A developer would need to know to change both spots or any other spots used.

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionService.java
```diff
@@ -114,4 +115,13 @@ public interface JobExecutionService {
    */
   Future<Boolean> completeJobExecutionWithError(String jobExecutionId, OkapiConnectionParams params);
 
+
+  /**
+   *
+   * @param ids JobExecution Logs to be deleted using Ids
```
okolawole-ebsco, line 121: I think "job execution logs" are referring to something different from "job executions". I think this should be `JobExecutions to be deleted using Ids`

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: Hi @okolawole-ebsco - I agree with your point. I feel we should have a generic builder for all types of queries.
We also get such libraries to use directly.
I will try to research the best fit for us.

I am thinking to go ahead with the merge and pick this change at a later stage, please share your thoughts.

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionService.java
SatishkumarPurswani: Hi @okolawole-ebsco - I have update the comments as per your suggestions, Thanks 🙂

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
okolawole-ebsco: yep, that is fine. The comment should suffice for now.

okolawole-ebsco (approved):

PR #622, opened 2022-04-21
MODSOURMAN-781 Populate dataImportEventPayload with correlationId value

viacheslavkol:
## Purpose
`correlationId` header if it exists should be passed to all data-import events. It's needed to track events that are related to the same action.

## Approach
- Populate dataImportPayload with correlationId value
- Update tests
- Increase data-import-processing-core version

## Learning
[MODSOURMAN-781](https://issues.folio.org/browse/MODSOURMAN-781)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,4 @@
-*
+#*
 !Dockerfile
 !docker*
 !mod-source-record-manager-server/target/*.jar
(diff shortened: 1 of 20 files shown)


viacheslavkol:
Commit [c640e56](https://github.com/folio-org/mod-source-record-manager/pull/622/commits/c640e5604f0de685f92714b9aa7cfbcc1c869760) must be reverted before merging

viacheslavkol:
It was decided to implement different approach on QuickMarc so this changes are no longer needed.

PR #624, opened 2022-04-26
MODSOURMAN-780 - Implement endpoint for adding summary for work accomplished in a job

RuslanLavrov:
## Purpose
to provide summary results of the import job for the summary section on the logs page

## Approach
* define endpoint `GET /metadata-provider/jobSummary/{jobExecutionId}`
* add migration script to create `get_job_execution_summary()` function for retrieving import job summary by jobExecution id
  * return 0 as total updated invoices to avoid redundant calculation since invoice update is not going to be supported
  * for invoice records summary should be calculated per invoice, not per invoice line
* add tests

## Learning
[MODSOURMAN-780](https://issues.folio.org/browse/MODSOURMAN-780)

##
can be merged after: https://github.com/folio-org/data-import-raml-storage/pull/255


mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -381,4 +425,37 @@ private org.folio.rest.jaxrs.model.ActionStatus mapNameToEntityActionStatus(Stri
   private JobLogEntryDto.SourceRecordType mapToEntityType(String entityType) {
     return entityType == null ? null : JobLogEntryDto.SourceRecordType.fromValue(entityType);
   }
+
+  private JobExecutionSummaryDto mapRowSetToJobExecutionSummaryDto(RowSet<Row> rowSet) {
+    JobExecutionSummaryDto jobExecutionSummaryDto = new JobExecutionSummaryDto();
+
+    rowSet.forEach(row -> jobExecutionSummaryDto
```
AivarIUsupov, line 432: Hi @RuslanLavrov as I understand, we'll only have one row, isn't it better to use rowSet.iterator().next() and make it as "return new JobExecutionSummaryDto.with.....".
It will make the code more clear.

SerhiiNosko (approved):

AivarIUsupov (approved):

PR #627, opened 2022-04-29
MODSOURMAN-779: Add status for Import jobs that are stopped by users.

VRohach:
## Purpose
Added new status "CANCELLED" for job that are stopped by users manually.

## Approach
Extended mechanism for status and ui-status execution

## Learning
https://issues.folio.org/browse/MODSOURMAN-779

PR #628, opened 2022-05-04
MODSOURMAN-779: Add status for Import jobs that are stopped by users.

VRohach:
## Purpose
Added new status "CANCELLED" for job that are stopped by users manually.

## Approach
Extended mechanism for status and ui-status execution

## Learning
https://issues.folio.org/browse/MODSOURMAN-779


VRohach:
This PR is the same as:https://github.com/folio-org/mod-source-record-manager/pull/627/files

But it seems like there are some issues on that branch, that's why I created this one.

RuslanLavrov (approved):

PR #629, opened 2022-05-04
MODSOURMAN-791 Reduce parsed record string conversion into MARC4J record

okolawole-ebsco:
## Purpose
In the AdditionalFieldsUtil class, there are numerous methods used to modify/introspect MARC records. Each of these methods coverts the parsed content represented as a JSON string into a MARC4J reader to generate a MARC4J record. This conversion is compute intensive. The aim is to reduce the conversion from parsed content string to MARC4J record.

## Approach
Cache MARC4J records with the originating parsed content string as its key. This cache process is not perfect but works well since any change, such as updating a subfield, generate a brand new parsed content string. Using the whole parsed content string as a key reduces the possibility of the cache returning stale data. The immutability of the String class in Java is relied upon to make this work.

The cache is limited to 25 items for now. That is, only 25 MARC4J records will be cached. New MARC4J records will evict old MARC4J records to limit memory overruns. The life span of a parsed content string within SRM is short. Caching this way should reduce computation by over 50%.

## Learning
https://issues.folio.org/browse/MODSOURMAN-791

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/JobMonitoringServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/JobMonitoringServiceImpl.java
@@ -1,6 +1,7 @@
 package org.folio.services;
 
 import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
 import java.util.Date;
 import java.util.List;
@@ -39,7 +40,7 @@ public class JobMonitoringServiceImpl implements JobMonitoringService {
 
   @Override
   public Future<List<JobMonitoring>> getInactiveJobMonitors(Long maxInactiveInterval, String tenantId) {
-    var lastInactiveDateTime = LocalDateTime.now().minus(maxInactiveInterval, ChronoUnit.MILLIS);
+    var lastInactiveDateTime = LocalDateTime.now(ZoneOffset.UTC).minus(maxInactiveInterval, ChronoUnit.MILLIS);
     return jobMonitoringDao.findByNotificationBeforeTimestamp(lastInactiveDateTime, false, tenantId);
   }
 
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
@@ -12,7 +12,6 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.folio.rest.jaxrs.model.Record;
 import org.marc4j.MarcJsonReader;
 import org.marc4j.MarcJsonWriter;
-import org.marc4j.MarcReader;
 import org.marc4j.MarcStreamWriter;
 import org.marc4j.MarcWriter;
 import org.marc4j.marc.ControlField;
@@ -26,7 +25,6 @@ import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Objects;
-import java.util.concurrent.ExecutionException;
 
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 
@@ -327,7 +325,8 @@ public final class AdditionalFieldsUtil {
 
   /**
    * Generate a {@link org.marc4j.marc.Record} from {@link Record} passed in.
-   * Will return null when there is no parsed content string present
+   * Will return null when there is no parsed content string present. Generated MARC record will be saved into cache if
+   * its parsed content string is not present in the cache as a key
    */
   private static org.marc4j.marc.Record computeMarcRecord(Record record) {
     if (record != null
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
@@ -62,7 +62,7 @@ public final class AdditionalFieldsUtil {
       };
 
     parsedRecordContentCache = CacheBuilder.newBuilder()
-      .maximumSize(25)
+      .maximumSize(2000)
       // weak keys allows parsed content strings that are used as keys to be garbage collected, even it is still
       // referenced by the cache.
       .weakKeys()
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
@@ -40,8 +40,8 @@ public final class AdditionalFieldsUtil {
 
   private static final Logger LOGGER = LogManager.getLogger();
 
-  private static CacheLoader<Object, org.marc4j.marc.Record> parsedRecordContentCacheLoader;
-  private static LoadingCache<Object, org.marc4j.marc.Record> parsedRecordContentCache;
+  private final static CacheLoader<Object, org.marc4j.marc.Record> parsedRecordContentCacheLoader;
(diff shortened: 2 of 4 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/JobMonitoringServiceImpl.java
```diff
@@ -39,7 +40,7 @@ public Future<List<JobMonitoring>> getAll(String tenantId) {
 
   @Override
   public Future<List<JobMonitoring>> getInactiveJobMonitors(Long maxInactiveInterval, String tenantId) {
-    var lastInactiveDateTime = LocalDateTime.now().minus(maxInactiveInterval, ChronoUnit.MILLIS);
+    var lastInactiveDateTime = LocalDateTime.now(ZoneOffset.UTC).minus(maxInactiveInterval, ChronoUnit.MILLIS);
```
okolawole-ebsco, line 43: I had to make this change due to failing tests in my local env, the PR branches but not in master for some weird reason.

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -262,43 +294,57 @@ public static String getValue(Record record, String tag, char subfield) {
...
-    return new MarcJsonReader(new ByteArrayInputStream(record.getParsedRecord().getContent().toString().getBytes(StandardCharsets.UTF_8)));
+  /**
+   * Generate a {@link org.marc4j.marc.Record} from {@link Record} passed in.
+   * Will return null when there is no parsed content string present. Generated MARC record will be saved into cache if
+   * its parsed content string is not present in the cache as a key
+   */
+  private static org.marc4j.marc.Record computeMarcRecord(Record record) {
+    if (record != null
+        && record.getParsedRecord() != null
+        && !StringUtils.isBlank(record.getParsedRecord().getContent().toString())) {
```
SerhiiNosko, line 334: We can simplify like this: record.getParsedRecord().getContent() != null

mod-source-record-manager-server/src/test/java/org/folio/services/afterprocessing/AdditionalFieldsUtilTest.java
```diff
@@ -285,4 +288,101 @@ public void shouldNotSortExistingFieldsWhenAddFieldToToMarcRecord() {
     Assert.assertTrue(added);
     Assert.assertEquals(expectedParsedContent, parsedRecord.getContent());
   }
+
+  @Test
```
SerhiiNosko, line 292: We can make separate tests for each scenario, in future if something change - I think it would be easier for readers to understand and modify particular test and not thinking about changing assert counters for other test cases

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
okolawole-ebsco: That is what i had prior. I wrote a test where the string was `""` and it failed so i made it more robust.

mod-source-record-manager-server/src/test/java/org/folio/services/afterprocessing/AdditionalFieldsUtilTest.java
SerhiiNosko: It makes sense

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -36,9 +40,46 @@ public final class AdditionalFieldsUtil {
...
+                parsedRecordContent.toString().getBytes(StandardCharsets.UTF_8)));
+          if (marcJsonReader.hasNext()) {
+            return marcJsonReader.next();
+          }
+          return null;
+        }
+      };
+
+    parsedRecordContentCache = CacheBuilder.newBuilder()
+      .maximumSize(2000)
```
SerhiiNosko, line 65: 2000 chosen based on average size of parsed record or some other things considered?
Here is our recommendation for memory allocation for modules:
image
More details here: https://wiki.folio.org/pages/viewpage.action?pageId=59932680

okolawole-ebsco: I had updated the JIRA with a comment on how i came to the current value. Here it is below

> After this story was implemented, some profiling was done to determine if the original cache size of 25 was adequate. While the cache was being utilized and overall the number of cache hits was greater than cache misses, increasing the cache size reduces the number of cache misses for an import of about 500 records. The cache miss count stabilized at about 1600. The new cache size to be used is 2000.

It is why i wanted to know what the maximum number of records in a chunk. Memory usage did not bump to significant level. The parsed content string will be garbage collected when the record(not MARC4J) is GC'ed. I believe when the key of the cache is gone, the value of the cache is removed eventually. So i think this value could be higher but i decided to draw a line here with my 500 import file.

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -52,12 +93,11 @@ public static boolean addFieldToMarcRecord(Record record, String field, char sub
     boolean result = false;
     try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
       if (record != null && record.getParsedRecord() != null && record.getParsedRecord().getContent() != null) {
-        MarcReader reader = buildMarcReader(record);
         MarcWriter streamWriter = new MarcStreamWriter(new ByteArrayOutputStream());
         MarcJsonWriter jsonWriter = new MarcJsonWriter(os);
         MarcFactory factory = MarcFactory.newInstance();
-        if (reader.hasNext()) {
-          org.marc4j.marc.Record marcRecord = reader.next();
+        org.marc4j.marc.Record marcRecord = parsedRecordContentCache.get(record.getParsedRecord().getContent());
```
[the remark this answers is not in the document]
okolawole-ebsco: good catch!

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -262,43 +306,57 @@ public static String getValue(Record record, String tag, char subfield) {
...
+          marcRecord.removeVariableField(variableField);
         }
+        // use stream writer to recalculate leader
+        marcStreamWriter.write(marcRecord);
+        marcJsonWriter.write(marcRecord);
+        record.setParsedRecord(
+            record
+                .getParsedRecord()
+                .withContent(new JsonObject(baos.toString()).encode()));
+        result = true;
```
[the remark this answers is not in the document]
okolawole-ebsco: I updated cache based on methods i saw where called constantly during my profiling. I am not against adding here. From what i can see, the removeField method is never called by anything other than tests. If you feel strongly, i can add it in.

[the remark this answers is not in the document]
SerhiiNosko: If its never called we can delete this method at all(in order to not support in future), if we going to remain for now - for me better to add cache here to be consistent.

SerhiiNosko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
SerhiiNosko: Good to know, thanks

Aliaksandr-Fedasiuk (approved):

PR #630, opened 2022-05-05
MODSOURMAN-785 - Filter out "Deleted" DI logs for GET JobExecutions API

SatishkumarPurswani:
## Purpose
Filter out "Deleted" DI logs for GET JobExecutions API

## Approach
APIs using the job_execution table will not return the deleted records. Deleted records are identified by the is_deleted column value being true.

#### TODOS and Open Questions
- [x] Will mark the check, once test cases are implemented.

### Marked checkboxes show changes implemented for specific APIs.
- [x] /metadata-provider/jobExecutions
- [x] /metadata-provider/journalRecords/{jobExecutionId}
- [x] /change-manager/jobExecutions/{id}
- [x] /change-manager/jobExecutions/{id}/children
- [ ] ~/metadata-provider/jobLogEntries/{jobExecutionId}~ - using get_job_log_entries table to fetch records, Hence no changes made.
- [ ] ~/metadata-provider/jobLogEntries/{jobExecutionId}/records/{recordId}~ - using get_record_processing_log table to fetch records, Hence no changes made.


mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -142,7 +142,7 @@ public Future<Optional<JobExecution>> getJobExecutionById(String id, String tena
     Promise<RowSet<Row>> promise = Promise.promise();
     try {
       String jobTable = formatFullTableName(tenantId, TABLE_NAME);
-      String query = format(GET_BY_ID_SQL, jobTable);
+      String query = format(GET_BY_ID_SQL, jobTable).concat("where is_deleted = false");
```
RuslanLavrov, line 145: Hello @SatishkumarPurswani, GET_BY_ID_SQL query template already contains `WHERE` clause, please add predefined condition `is_deleted = false` to GET_BY_ID_SQL template to fix the query execution. Thank you!

okolawole-ebsco:
Any update to NEWS.md?

okolawole-ebsco (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionFilter.java
```diff
@@ -31,6 +32,13 @@ public class JobExecutionFilter {
   private String userId;
   private Date completedAfter;
   private Date completedBefore;
+  private boolean isDeleted;
+
+
+  public JobExecutionFilter isDeleted(boolean isDeleted) {
+    this.isDeleted=isDeleted;
```
SerhiiNosko, line 39: missing some spaces here: this.isDeleted = isDeleted;

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionFilter.java
```diff
@@ -127,10 +135,17 @@ public String buildCriteria() {
     if (completedBefore != null) {
       addCondition(conditionBuilder, buildLessThanOrEqualCondition(COMPLETED_DATE_FIELD, formatter.format(completedBefore)));
     }
+    setIsDeletedCondition(conditionBuilder);
     addCondition(conditionBuilder, buildNotBoolCondition(JOB_PROFILE_HIDDEN_FIELD));
     return conditionBuilder.toString();
   }
 
+  private void setIsDeletedCondition(StringBuilder conditionBuilder) {
+    if(!isDeleted){
```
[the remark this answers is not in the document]
SerhiiNosko: Or even better to do the same as for hidden, so reuse this method: buildNotBoolCondition(JOB_PROFILE_HIDDEN_FIELD)
Because we don't need to show only deleted, so to simplify can follow approach with is_profile_hidden field

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/metadataProvider/MetadataProviderJobExecutionAPITest.java
```diff
@@ -1044,4 +1046,57 @@ private Future<JournalRecord> createJournalRecord(String jobExecutionId, String
...
+  @Test
+  public void shouldNotReturnDeletedJobExecutionRecords(TestContext context) {
+    Async async = context.async();
+    JobExecution createdJobExecution = constructAndPostInitJobExecutionRqDto(1).getJobExecutions().get(0);
+    String sourceRecordId = UUID.randomUUID().toString();
+    String recordTitle = "test title";
+
+    Future<JournalRecord> future = Future.succeededFuture()
+      .compose(v -> createJournalRecord(createdJobExecution.getId(), sourceRecordId, null, null, recordTitle, 0, CREATE, MARC_BIBLIOGRAPHIC, COMPLETED, null))
+      .compose(v -> createJournalRecord(createdJobExecution.getId(), sourceRecordId, null, null, null, 0, UPDATE, MARC_BIBLIOGRAPHIC, COMPLETED, null))
```
RuslanLavrov, line 1059: Hello @SatishkumarPurswani, looks like we don't need to create journal records because below in the test we don't check that they should not be returned

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/changeManager/ChangeManagerAPITest.java
```diff
@@ -2000,4 +2001,93 @@ public void testDeleteChangeManagerJobExecutionsIncorrectBody(){
...
+      .then().statusCode(HttpStatus.SC_CREATED)
+      .extract().path("parentJobExecutionId");
+
+    // when
+    String parentJobExecutionId_2 = RestAssured.given()
+      .spec(spec)
+      .body(JsonObject.mapFrom(requestDto).toString())
+      .when().post(JOB_EXECUTION_PATH)
+      .then().statusCode(HttpStatus.SC_CREATED)
+      .extract().path("parentJobExecutionId");
```
RuslanLavrov, line 2039:
```suggestion
    String parentJobExecutionId = constructAndPostInitJobExecutionRqDto(1).getParentJobExecutionId();
    String parentJobExecutionId_2 = constructAndPostInitJobExecutionRqDto(1).getParentJobExecutionId();
```
propose to simplify job executions creation same as in the MetadataProviderJobExecutionAPITest.shouldNotReturnDeletedJobExecutionRecords() test

SerhiiNosko (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionFilter.java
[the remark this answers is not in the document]
SatishkumarPurswani: Hi @SerhiiNosko - code has been updated as per your suggestions, Thank you 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionFilter.java
SatishkumarPurswani: Hi @SerhiiNosko - This code was not required, Hence removed. Thank you 🙂

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/metadataProvider/MetadataProviderJobExecutionAPITest.java
SatishkumarPurswani: Hi @RuslanLavrov - code has been updated as per your suggestions, Thank you 🙂

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/changeManager/ChangeManagerAPITest.java
SatishkumarPurswani: Hi @RuslanLavrov - code has been updated as per your suggestions, Thank you 🙂

RuslanLavrov: thank you!

Abdulkhakimov (approved):

PR #636, opened 2022-05-12
MODSOURMAN-794 - Removed API /metadata-provider/logs/{jobExecutionId} and related files

SatishkumarPurswani:
## Purpose
API - /metadata-provider/logs/{jobExecutionId} was not required and was not being used.

## Approach
Removed the APIs related code, test cases, and files

## Learning
Have created this PR irrespective of MODSOURMAN-785 as the change are not much related to the same Jira but can only be referenced.


mod-source-record-manager-server/src/test/java/org/folio/verticle/consumers/DataImportJournalConsumerVerticleTest.java
```diff
@@ -101,167 +95,6 @@ public void setUp() {
...
-        .withContentType(ACTION_PROFILE)
-        .withContent(JsonObject.mapFrom(new ActionProfile().withFolioRecord(ActionProfile.FolioRecord.HOLDINGS))))
-      .withEventsChain(List.of(DI_SRS_MARC_BIB_RECORD_CREATED.value(), DI_INVENTORY_HOLDING_CREATED.value()));
-
-    // when
-    KafkaConsumerRecord<String, String> kafkaConsumerRecord = buildKafkaConsumerRecord(eventPayload);
-    dataImportJournalKafkaHandler.handle(kafkaConsumerRecord);
-
-    // then
-    Future<JobExecutionLogDto> future = journalRecordDao.getJobExecutionLogDto(jobExecution.getId(), TENANT_ID);
```
SerhiiNosko, line 257: I think we don't need to delete these test methods, just need to replace getJobExecutionLogDto() to getByJobExecutionId()

okolawole-ebsco:
Why is this endpoint being removed? what endpoint do we use to retrieve a log after this is removed? or we dont retrieve logs anymore?

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/test/java/org/folio/verticle/consumers/DataImportJournalConsumerVerticleTest.java
SatishkumarPurswani: Hi @SerhiiNosko - Updated, Thank you 🙂

SatishkumarPurswani:
> Why is this endpoint being removed? what endpoint do we use to retrieve a log after this is removed? or we dont retrieve logs anymore?

Hi @okolawole-ebsco - 
API - /metadata-provider/logs/{jobExecutionId} is not required and was not being used.
We have another API to retrieve logs - /change-manager/jobExecutions/{id}

okolawole-ebsco:
> > Why is this endpoint being removed? what endpoint do we use to retrieve a log after this is removed? or we dont retrieve logs anymore?
> 
> Hi @okolawole-ebsco - API - /metadata-provider/logs/{jobExecutionId} is not required and was not being used. We have another API to retrieve logs - /change-manager/jobExecutions/{id}

Thanks! was this endpoint added in the release? if it was added a lot earlier, then this will constitute a contract change. I think product management would need to be notified to take next steps.

okolawole-ebsco (approved):

PR #639, opened 2022-05-16
MODSOURMAN-789 Implement endpoint to get job executions profiles

Abdulkhakimov:
## Purpose
Implement new endpoint for retrieving information about related job profiles based on job executions

#### TODOS and Open Questions
- [x] Define endpoint to retrieve a list of unique jobProfileInfo elements.
- [x] The endpoint should provide pagination query parameters.
- [x] The endpoint should require "metadata-provider.jobexecutions.get" permission necessary for jobExecution retrieving.
- [x] Implement a method to Dao component to retrieve list of jobProfileInfo elements based on data from "job_execution" table.
- [x] Add tests.

## Learning
[MODSOURMAN-789](https://issues.folio.org/browse/MODSOURMAN-789)


mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -201,6 +202,11 @@ public Future<JobExecution> setJobProfileToJobExecution(String jobExecutionId, J
         }));
   }
 
+  @Override
+  public Future<JobProfileInfoCollection> getRelatedJobProfiles(int offset, int limit, String tenantId) {
+    return jobExecutionDao.getRelatedJobProfiles(offset, limit, tenantId);
+  }
+
```
SerhiiNosko, line 209: This endpoint necessary for building dropdown with all possible job profiles to be able to filter by them.
Not sure that UI team will do paginated requests to fetch all job profiles. Thinking that we don't need pagination here

Abdulkhakimov: Hi @SerhiiNosko. It was specified in the ticket. If reporter(Ruslan Lavrov) is agree with you then I can remove offset param.

RuslanLavrov (approved):

SatishkumarPurswani (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -286,10 +303,16 @@ private Tuple mapToTuple(JobExecution jobExecution) {
...
+    rowSet.iterator().forEachRemaining(row -> {
       jobCollection.getJobExecutions().add(mapRowToJobExecutionDto(row));
       jobCollection.setTotalRecords(row.getInteger(TOTAL_COUNT_FIELD));
-    }
+    });
+    return jobCollection;
+  }
+
+  private JobProfileInfoCollection mapRowToJobProfileInfoCollection(RowSet<Row> rowSet) {
+    JobProfileInfoCollection jobCollection = new JobProfileInfoCollection().withTotalRecords(rowSet.rowCount());
```
SerhiiNosko, line 314: rowSet.rowCount return total number of unique profiles, not only returned in response for current limit, offset, is it correct?

Abdulkhakimov: Hi @SerhiiNosko. unfortunately, only the number including offset and limit.

SerhiiNosko: Basically its not correct contract for pagination, because consumers of the paginated endpoint need to know how many requests need to done to fetch all records.
Can you please check how pagination implemented in other endpoints?

Abdulkhakimov: I hadn't considered that point. It will be fixed. Thank you)

SerhiiNosko (approved):

PR #643, opened 2022-05-18
MODSOURMAN-790: Implement endpoint to get job executions users.

VRohach:
## Purpose
Create a brand-new endpoint for retrieving unique users (with userId, firstName, lastName fields) who created job executions.

## Approach
* New endpoint created.
* Tests added

## Learning
JIRA: https://issues.folio.org/browse/MODSOURMAN-790

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
@@ -173,7 +173,7 @@ public class MetadataProviderImpl implements MetadataProvider {
   public void getMetadataProviderJobExecutionsUsers(int offset, int limit, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     vertxContext.runOnContext(v -> {
       try {
-        jobExecutionService.getUniqueUsersInfo(offset, limit, tenantId)
+        jobExecutionService.getRelatedUsersInfo(offset, limit, tenantId)
           .map(GetMetadataProviderJobExecutionsUsersResponse::respond200WithApplicationJson)
           .map(Response.class::cast)
           .otherwise(ExceptionHelper::mapExceptionToResponse)
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
@@ -100,5 +100,5 @@ public interface JobExecutionDao {
    * @param limit    maximum number of results to return
    * @return collection of JobExecutionUserInfoCollection dtos with userIds, firstNames, lastNames
    */
-  Future<JobExecutionUserInfoCollection> getUniqueUsersInfo(int offset, int limit, String tenantId);
+  Future<JobExecutionUserInfoCollection> getRelatedUsersInfo(int offset, int limit, String tenantId);
 }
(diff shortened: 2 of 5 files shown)


mod-source-record-manager-server/src/main/java/org/folio/dao/util/JobExecutionDBConstants.java
```diff
@@ -80,6 +82,11 @@ public final class JobExecutionDBConstants {
     "LEFT JOIN total p ON true " +
     "LIMIT $1 OFFSET $2";
 
+  public static final String GET_UNIQUE_USERS = "WITH unique_users AS (SELECT DISTINCT user_id, " +
+    "job_user_first_name, job_user_last_name FROM %s)," +
+    "total AS (SELECT count(*) AS total_count FROM unique_users)" +
+    "SELECT j.*, p.* FROM unique_users j LEFT JOIN total p ON true LIMIT $1 OFFSET $2";
+
```
SerhiiNosko, line 89: Seems it also required to filter by is_hidden field here, because delete authority job uses hidden profile, and user for it should not be provided in response

VRohach: Hi @SerhiiNosko! Do you mean 'is_deleted'? I added a condition for it.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
```diff
@@ -169,6 +169,22 @@ public void getMetadataProviderJobProfiles(int offset, int limit, Map<String, St
...
   }
 
+  @Override
+  public void getMetadataProviderJobExecutionsUniqueUsers(int offset, int limit, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    vertxContext.runOnContext(v -> {
+      try {
+        jobExecutionService.getUniqueUsersInfo(offset, limit, tenantId)
+          .map(GetMetadataProviderJobExecutionsUniqueUsersResponse::respond200WithApplicationJson)
+          .map(Response.class::cast)
+          .otherwise(ExceptionHelper::mapExceptionToResponse)
```
okolawole-ebsco, line 179: when an exception occurs, would it be written to the logs?

mod-source-record-manager-server/src/main/java/org/folio/dao/util/JobExecutionDBConstants.java
SerhiiNosko: I mean job_profile_hidden, the same as in related endpoint for another dropdown:
https://github.com/folio-org/mod-source-record-manager/commit/e1b2bf7ecf2e4ed8f48f325546db1c68f64c9656#diff-426ed7c82b47e1e393acd421afd7c05b73d48f13cdb3e2fae4afb5d024b72f10R76

VRohach: @SerhiiNosko I added new condition!

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
VRohach: Hi @okolawole-ebsco! Yes, this is a standard mechanism for exception catching+logging.

okolawole-ebsco (approved):

SerhiiNosko (approved):

Abdulkhakimov (approved):

RuslanLavrov (approved):

PR #646, opened 2022-05-24
MODSOURMAN-787 - Code changes for deleting Job Execution periodically.

SatishkumarPurswani:
## Purpose
This code will be executed every 24hours to delete the JobExecutions records already marked for deletion.
Deletion will take places on these tables - job_execution_progress, job_monitoring, job_execution_source_chunks, journal_records, job_execution.

## Approach
We are fetching the records marked for deletion every 24hours and deleting them from these tables - job_execution_progress, job_monitoring, job_execution_source_chunks, journal_records, job_execution.

We have a configurable property for running periodic job execution with a default value set to 24hours, Here is the property name:
periodic.job.execution.permanent.delete.interval.ms = 86400000

We have a configurable property to check the difference between the completed_date and current date, with a default value set to 2days, Here is the property name:
job.execution.difference.number.of.days = 2

#### TODOS and Open Questions
- [x] Test cases added.
- [x] Tested code locally.

## Learning
https://issues.folio.org/browse/MODSOURMAN-787


mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -109,6 +110,13 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
   public static final String TENANT_NAME = "tenantName";
   public static final String TRUE = "true";
   public static final String DB_TABLE_NAME_FIELD = "tableName";
+  public static final String SELECT_ID_WHERE_IS_DELETED_TRUE = "SELECT id FROM %s WHERE is_deleted = true";
```
SerhiiNosko, line 113: We need to include completed_date in SELECT query

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -109,6 +110,13 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
...
+  public static final String DELETE_QUERY = "DELETE from %s where %s IN ('%s')";
+  public static final String JOB_MONITORING = "job_monitoring";
+  public static final String JOB_EXECUTION_SOURCE_CHUNKS = "job_execution_source_chunks";
+  public static final String JOURNAL_RECORDS = "journal_records";
+  public static final String JOB_EXECUTION_ID = "job_execution_id";
+  public static final String JOBEXECUTIONID = "jobexecutionid";
```
SerhiiNosko, line 119: Its needed to be refactored in some way to avoid to identical by meaning constant names: JOB_EXECUTION_ID, JOBEXECUTIONID

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
```diff
@@ -101,4 +101,11 @@ public interface JobExecutionDao {
    * @return collection of JobExecutionUserInfoCollection dtos with userIds, firstNames, lastNames
    */
   Future<JobExecutionUserInfoCollection> getRelatedUsersInfo(int offset, int limit, String tenantId);
+
+  /**
```
SerhiiNosko, line 105: Need to update java docs with some words what method doing

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -451,4 +459,62 @@ private String buildOrderByClause(List<SortField> sortFields) {
...
   }
 
+  @Override
+  public boolean permanentDeleteJobExecutions(String tenantName) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    Promise<SQLConnection> connection = Promise.promise();
+    String rollbackMessage = "Rollback transaction. Error during jobExecution deletion process.";
+    Future.succeededFuture()
+      .compose(c -> {
+        pgClientFactory.createInstance(tenantName).startTx(connection);
```
SerhiiNosko, line 469: We can move statement: pgClientFactory.createInstance(tenantName) to upper level and not create it each time something like this:
PostgresClient pgClient = pgClientFactory.createInstance(tenantId);

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -451,4 +459,62 @@ private String buildOrderByClause(List<SortField> sortFields) {
...
+        pgClientFactory.createInstance(tenantName).endTx(connection.future(), endTxFuture);
+        return endTxFuture.future();
+      }).onComplete(ar -> {
+        if (ar.failed()) {
+          pgClientFactory.createInstance(tenantName).rollbackTx(connection.future(), rollback -> promise.fail(ar.cause()));
+          return;
+        }
+        promise.complete();
+      });
+    return promise.tryComplete();
```
SerhiiNosko, line 498: Basically it mismatches java doc contract for this method  in some way, instead of saying that deletion competed with errors or not - it says does deletion process completed or still in progress

mod-source-record-manager-server/src/main/java/org/folio/verticle/DeleteJobExecutionPeriodicallyVerticle.java
```diff
@@ -0,0 +1,63 @@
...
+  private static AbstractApplicationContext springGlobalContext;
+
+  @Autowired
+  private JobExecutionDao jobExecutionDao;
+
+  @Value("${job.execution.permanent.delete.interval.ms:86400000}")
+  private long jobExecutionPermanentDeletionInterval;
+  private long timerId;
+
+  public static void setSpringContext(AbstractApplicationContext springContext) {
```
SerhiiNosko, line 28: We already have in SRM 2 verticles that doing some periodical jobs and it will be the third verticle. We can create new abstract class with name something like this AbstractPeriodicallyJobsVerticle and move all this spring stuff to abstract class. It will be easier to refactor in future when we change approach of initiatialization of spring context.
And we will store common things connected with periodically execution in 1 abstract class.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: Hi @SerhiiNosko - Updated code as per your suggestion. Thank you 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
SatishkumarPurswani: Hi @SerhiiNosko - Updated code as per your suggestion. Thank you 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: Hi @SerhiiNosko - Updated code as per your suggestion. Thank you 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SatishkumarPurswani: Hi @SerhiiNosko - Updated code as per your suggestion. Thank you 🙂

mod-source-record-manager-server/src/main/java/org/folio/dao/util/TenantUtil.java
```diff
@@ -0,0 +1,40 @@
...
+
+/* The util class helps to work with tenant's data */
+public class TenantUtil {
+
+  private TenantUtil() {
+  }
+
+
+  /* The method returns all the available tenants for whom the mode is deployed */
+  public static Future<List<String>> getModuleTenants(Vertx vertx) {
```
SerhiiNosko, line 22: It should not be utility class, it should be common spring service class, that has vertx as param in constructor, working on this change

mod-source-record-manager-server/src/main/java/org/folio/verticle/DeleteJobExecutionPeriodicallyVerticle.java
SerhiiNosko: I made this refactoring, so resolving the comment

mod-source-record-manager-server/src/main/java/org/folio/dao/util/TenantUtil.java
SerhiiNosko: Moved from utility class to spring based component

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
```diff
@@ -101,4 +103,14 @@ public interface JobExecutionDao {
...
+   * Permanently deletes Job Executions from related tables depending upon difference in number of days
+   *
+   * @param tenantName tenant name
+   * @param diffNumberOfDays difference in Number of Days from the day record marked for deletion
+   * @return future with RowSet information
+   */
+
+  Future<RowSet<Row>> hardDeleteJobExecutions(String tenantName, long diffNumberOfDays);
```
okolawole-ebsco, line 115: Question, why is a set of rows being returned? I would expect may be a `void` or boolean to signify that the hard delete was successful.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -451,4 +460,64 @@ private String buildOrderByClause(List<SortField> sortFields) {
       .collect(Collectors.joining(", ", "ORDER BY ", EMPTY));
   }
 
+  @Override
+  public Future<RowSet<Row>> hardDeleteJobExecutions(String tenantName, long diffNumberOfDays) {
+    Promise<RowSet<Row>> promise = Promise.promise();
+    Promise<SQLConnection> connection = Promise.promise();
+    PostgresClient postgresClient = pgClientFactory.createInstance(tenantName);
+    String rollbackMessage = "Rollback transaction. Error during jobExecution deletion process.";
+    Future.succeededFuture()
```
okolawole-ebsco: nitpick: thinking you could just start the async chain with `postgresClient.startTx` rather than `succeededFuture`. You might have a reason for starting from a `succeededFuture`.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -451,4 +460,64 @@ private String buildOrderByClause(List<SortField> sortFields) {
...
+      .compose(c -> {
+        postgresClient.startTx(connection);
+        return connection.future();
+      })
+      .compose(s -> fetchJobExecutionIdsToBeConsideredForDeleting(tenantName, diffNumberOfDays, connection, postgresClient))
+      .compose(rowSet -> {
+        if (rowSet.rowCount() < 1) {
+          throw new NotFoundException(rollbackMessage);
```
okolawole-ebsco: I think it is possible for this process to run and not find any jobs to hard delete. What do you think?

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
SerhiiNosko: I updated this method to return CompositeFuture, we returning future instead of void to be able to know when it is completed to write logs about results into periodical deletion job class. Also I've refactored approach to use transaction's wrapper from utility class.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -451,4 +460,64 @@ private String buildOrderByClause(List<SortField> sortFields) {
...
+        }
+        return mapRowsetValuesToListOfString(rowSet);
+      })
+      .compose(strings -> {
```
okolawole-ebsco: I think these tables need to be deleted in a specific order due to foreign key constraints. If not then maybe delete from the "top" should cascade all the way down to child tables, which would mean just one delete statement instead of many. I definitely see the benefit of deleting child tables first. What do you think?

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -451,4 +460,64 @@ private String buildOrderByClause(List<SortField> sortFields) {
...
+        Future<RowSet<Row>> jobMonitoringFuture = Future.future(rowSetPromise -> deleteJobExecutionsFromParticularTable(tenantName, DELETE_QUERY, connection, JOB_MONITORING, jobExecutionIds, JOB_EXECUTION_ID, rowSetPromise, postgresClient));
+        Future<RowSet<Row>> jobExecutionSourceChunksFuture = Future.future(rowSetPromise -> deleteJobExecutionsFromParticularTable(tenantName, DELETE_QUERY, connection, JOB_EXECUTION_SOURCE_CHUNKS, jobExecutionIds, JOBEXECUTIONID, rowSetPromise, postgresClient));
+        Future<RowSet<Row>> journalRecordsFuture = Future.future(rowSetPromise -> deleteJobExecutionsFromParticularTable(tenantName, DELETE_QUERY, connection, JOURNAL_RECORDS, jobExecutionIds, JOB_EXECUTION_ID, rowSetPromise, postgresClient));
+        Future<RowSet<Row>> jobExecutionFuture = Future.future(rowSetPromise -> deleteJobExecutionsFromParticularTable(tenantName, DELETE_QUERY, connection, TABLE_NAME, jobExecutionIds, ID, rowSetPromise, postgresClient));
```
okolawole-ebsco: nitpick: I would really like a builder pattern used for the sql queries generated. But you dont have much time left with us @SatishkumarPurswani 😄

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SerhiiNosko: Yes it is possible, approach has been changed without throwing exception(now we use utility class to wrap into transaction, so we could not fail future or throw exception, because it will be considered as action to rollback transaction by this utility. So now we rely on CollectionUtils.isEmpty() checking.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SerhiiNosko: This method has been refactored

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SerhiiNosko: Data from the main table job_execution are deleting in the last turn. During testing we did not observe any foreign key constraints, data from all 5 tables successfully deleted. But if you think we need to change order for some tables - let us know and we will retest it.

SerhiiNosko: Looking one more time, theoretically future to delete data from job_execution table can occur earlier, because its Composite Future, and Im not sure that order passing to Composite future is equals to order of future's execution, so potentially it can lead to Foreign Keys constraints, need to think

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -109,6 +106,13 @@ public class JobExecutionDaoImpl implements JobExecutionDao {
   public static final String TENANT_NAME = "tenantName";
   public static final String TRUE = "true";
   public static final String DB_TABLE_NAME_FIELD = "tableName";
+  public static final String SELECT_ID_WHERE_IS_DELETED_TRUE = "SELECT id FROM %s WHERE is_deleted = true and completed_date<= '%s'";
+  public static final String DELETE_QUERY = "DELETE from %s where %s IN ('%s')";
```
okolawole-ebsco, line 110: As mentioned before, I am not fond of this way of writing sql...

strings in the `IN` clause need to be parameterized. In its current state, the db will need to parse every sql statement because it will appear new. This comment is applicable to the SELECT statement above as well.

mod-source-record-manager-server/src/main/java/org/folio/services/TenantDataProviderImpl.java
```diff
@@ -0,0 +1,44 @@
...
+  @Autowired
+  public TenantDataProviderImpl(Vertx vertx) {
+    this.vertx = vertx;
+  }
+
+  @Override
+  public Future<List<String>> getModuleTenants() {
+    PostgresClient pgClient = PostgresClient.getInstance(vertx);
+    Promise<RowSet<Row>> promise = Promise.promise();
+    String tenantQuery = "select nspname from pg_catalog.pg_namespace where nspname LIKE '%_mod_source_record_manager';";
```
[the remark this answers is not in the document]
okolawole-ebsco: Thinking a call to okapi might be a better fit here.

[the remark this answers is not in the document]
SerhiiNosko: Its needed to create some SYSTEM user, that firstly need to authenticate with some creds, and after this we need to use token to get this info from Okapi.
So due to this complexity we using this approach with DB

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/AbstractPeriodicJobVerticle.java
```diff
@@ -0,0 +1,32 @@
...
+  private static AbstractApplicationContext springGlobalContext;
+
+  //TODO: get rid of this workaround with global spring context
+  public static void setSpringGlobalContext(AbstractApplicationContext springGlobalContext) {
+    AbstractPeriodicJobVerticle.springGlobalContext = springGlobalContext;
+  }
+
+  @Override
+  public void start() {
+    context.put("springContext", springGlobalContext);
```
okolawole-ebsco, line 22: I have seen this action a couple of times already. If `SprintContextUtil.autowireDependencies` wont work if there isn't an item called `springContext`, then i think this put should be wrapped around some other method. Success would mean that a user of `SpringContextUtil` would not need to know that `springContext` needs to be inserted into the vertx context.

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/PeriodicDeleteJobExecutionVerticle.java
```diff
@@ -0,0 +1,54 @@
...
+  @Autowired
+  private JobExecutionDao jobExecutionDao;
+  @Autowired
+  private TenantDataProvider tenantDataProvider;
+
+  private long timerId;
+
+  @Override
+  public void startPeriodicJob() {
+    timerId = vertx.setPeriodic(permanentDeletionInterval, handler -> proceedForJobExecutionPermanentDeletion());
```
okolawole-ebsco, line 35: Thinking this call to `setPeriodic` would be in `AbstractPeriodicJobVerticle`. As it stands there is nothing that makes `AbstractPeriodicJobVerticle` "periodic". It is possible to implement a non-periodic verticle.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SerhiiNosko: Refactored to use JOB_EXECUTION_ID and JOB_EXECUTION_ID_OLD_FORMAT constants

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SerhiiNosko: Refactored to delete firstly from 4 related tables and after this from main job_execution table. This approach will prevent foreign key constraint exceptions.

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/PeriodicDeleteJobExecutionVerticle.java
SerhiiNosko: Refactored abstract class, moved all periodic related code to abstract class

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/PeriodicDeleteJobExecutionVerticle.java
```diff
@@ -0,0 +1,54 @@
...
+  public void stop() throws Exception {
+    vertx.cancelTimer(timerId);
+    super.stop();
+  }
+
+  private void proceedForJobExecutionPermanentDeletion() {
+    tenantDataProvider.getModuleTenants()
+      .onSuccess(allTenants -> allTenants.forEach(tenantName -> {
+        log.info("Check tenant [{}] for marked as deleted jobs", tenantName);
+        jobExecutionDao.hardDeleteJobExecutions(tenantName, diffNumberOfDays)
```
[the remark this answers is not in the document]
SerhiiNosko: Changed signature to return Future<Boolean>

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/AbstractPeriodicJobVerticle.java
SerhiiNosko: Refactored approach of initializing verticles. Now they initialized as spring beans. New SpringVerticleFactory has been introduced.

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
SerhiiNosko: Changed both queries to be parameterized, thanks

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
```diff
@@ -32,15 +34,10 @@
...
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
-import java.util.Objects;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
```
Abdulkhakimov, line 40: Please remove wildcard import

SerhiiNosko: Removed, thanks

mod-source-record-manager-server/src/test/java/org/folio/dao/JobExecutionDaoImplTest.java
```diff
@@ -104,11 +144,131 @@ private Future<RowSet<Row>> createProgressForJobExecutions(List<JobExecution> jo
...
           .withCurrentlyFailed(random.nextInt(10))
           .withTotal(random.nextInt(10));
-
         future = future.compose(ar -> jobExecutionProgressDao.save(progress, params.getTenantId()));
       }
     }
     return future;
   }
 
+  @Test
```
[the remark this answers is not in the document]
SerhiiNosko: Refactoring has been completed

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDaoImpl.java
okolawole-ebsco: looks great!

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportConsumersVerticle.java
```diff
@@ -7,9 +7,20 @@
...
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_COMPLETED;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_ERROR;
-
+import static org.springframework.beans.factory.config.BeanDefinition.SCOPE_PROTOTYPE;
+
+/**
+ * Verticle to process DI_COMPLETE, DI_ERROR events.
```
okolawole-ebsco, line 18: awesome comment!

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/AbstractPeriodicJobVerticle.java
okolawole-ebsco: The refactored approach is lovely.

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/PeriodicDeleteJobExecutionVerticle.java
okolawole-ebsco: looks great!

mod-source-record-manager-server/src/main/java/org/folio/verticle/periodic/PeriodicDeleteJobExecutionVerticle.java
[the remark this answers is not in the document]
okolawole-ebsco: looks good.

mod-source-record-manager-server/src/test/java/org/folio/dao/JobExecutionDaoImplTest.java
```diff
@@ -104,11 +147,193 @@ private Future<RowSet<Row>> createProgressForJobExecutions(List<JobExecution> jo
...
+  public void shouldDeleteRecordsMarkedForJobExecutionsDeletion(TestContext context) {
+    /*
+      In this test we setup job execution that finishes 3 days ago, after this job execution has been soft deleted.
+      Periodic job should hard delete it, because it checks soft deleted jobs older than 2 days.
+    */
+    Async async = context.async();
+
+    Instant minusThreeDays = LocalDate.now().minusDays(3).atStartOfDay().atZone(ZoneId.systemDefault()).toInstant();
```
okolawole-ebsco, line 164: I think all the timezones in the test should be set to UTC specifically. This will allow developer machines to have accurate tests/ data in the db since most will not be configured for UTC.

mod-source-record-manager-server/src/test/java/org/folio/dao/JobExecutionDaoImplTest.java
```diff
@@ -104,11 +147,193 @@ private Future<RowSet<Row>> createProgressForJobExecutions(List<JobExecution> jo
...
+      .withSourceRecordOrder(0)
+      .withSourceId(UUID.randomUUID().toString())
+      .withEntityType(JournalRecord.EntityType.MARC_BIBLIOGRAPHIC)
+      .withEntityId(UUID.randomUUID().toString())
+      .withActionType(CREATE)
+      .withActionDate(new Date())
+      .withActionStatus(COMPLETED);
+
+    JobExecutionSourceChunk jobExecutionSourceChunk = new JobExecutionSourceChunk()
+      .withId("67dfac11-1caf-4470-9ad1-d533f6360bdd")
```
okolawole-ebsco, line 281: Does this id have to be a specific string or can `UUID.randomUUID()` work?

mod-source-record-manager-server/src/test/java/org/folio/dao/JobExecutionDaoImplTest.java
```diff
@@ -1,39 +1,70 @@
...
 import static org.folio.rest.util.OkapiConnectionParams.OKAPI_TENANT_HEADER;
 import static org.folio.rest.util.OkapiConnectionParams.OKAPI_TOKEN_HEADER;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 @RunWith(VertxUnitRunner.class)
+@Log4j2
 public class JobExecutionDaoImplTest extends AbstractRestTest {
+  public static final String GENERIC_SELECT_QUERY_TO_GET_COUNT = "select count(*) from %s where %s IN ('%s')";
```
okolawole-ebsco, line 65: nitpick: it is not as critical since this isn't application code, this could be parameterized as well.

mod-source-record-manager-server/src/test/java/org/folio/dao/JobExecutionDaoImplTest.java
SerhiiNosko: Missed it during development, updated to UTC, thanks

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionDao.java
okolawole-ebsco: The final revision of this has been set to `Future<Boolean>`

okolawole-ebsco (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportConsumersVerticle.java
SerhiiNosko: Thank you

VRohach (approved):

PR #649, opened 2022-05-25
MODSOURMAN-796. Change logic to initialize job execution progress after reading file instead of processing the first chunk

SerhiiNosko:
## Purpose
Due to flow control changes,  offsets of topic DI_RAW_RECORDS_CHUNK_READ consumed by batches. If next import triggers - it will not appear in Running section on UI, because job execution progress initialized after reading first chunk of DI_RAW_RECORDS_CHUNK_READ topic, so in order to see new import on UI - need to wait until all chunks of previous imports processed.

## Approach

- Introduce new event DI_INITIALIZATION_STARTED (Corresponding PR: https://github.com/folio-org/data-import-raml-storage/pull/263)
- Produce this new event from mod-data-import after finding out how many records input file contains (Corresponding PR: https://github.com/folio-org/mod-data-import/pull/217)
- Initialize job execution progress and update job execution status to be visible in Running section on UI before receiving the first chunk of DI_RAW_RECORDS_CHUNK_READ , this operation should be idempotent(current PR). 

## Learning
https://issues.folio.org/browse/MODSOURMAN-796

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
@@ -59,7 +59,7 @@ public class RawRecordsFlowControlServiceImpl implements RawRecordsFlowControlSe
    * threshold can be missed during resetting that can cause that resume/pause cycle may not be as usual, because we are
    * starting from clear state after reset, but any events would not be missed and consumers never pause forever.
    */
-  @Scheduled(cron = "${di.flow.control.reset.state.cron:0 0/5 * * * ?}")
+  @Scheduled(fixedRateString = "${di.flow.control.reset.state.interval:PT5M}")
   public void resetState() {
     if (!enableFlowControl) {
       return;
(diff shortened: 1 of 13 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -59,7 +59,7 @@ public void init() {
    * threshold can be missed during resetting that can cause that resume/pause cycle may not be as usual, because we are
    * starting from clear state after reset, but any events would not be missed and consumers never pause forever.
    */
-  @Scheduled(cron = "${di.flow.control.reset.state.cron:0 0/5 * * * ?}")
+  @Scheduled(fixedRateString = "${di.flow.control.reset.state.interval:PT5M}")
```
SerhiiNosko, line 62: I made this change, because its hard to setup cron expression in rancher, cron is cutting of in rancher, so basically this style is more convenient, now this job also triggered each 5 mins.
Here is example of syntacs to specifying values, its Spring based Duration class:
image

VRohach (approved):

PR #652, opened 2022-06-01
MODSOURMAN-802: Block sending "Cancel" signal to finished task.

VRohach:
## Purpose
To avoid error for the user when cancel job. It can be cause form time to time if the running job is already has 'commited' status.

## Approach
Check whether current job already has status 'commited' - skip next logic and return 204 OK status for user.

## Learning
https://issues.folio.org/browse/MODSOURMAN-802


mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -245,15 +248,21 @@ public Future<Boolean> completeJobExecutionWithError(String jobExecutionId, Okap
     return jobExecutionDao.getJobExecutionById(jobExecutionId, params.getTenantId())
       .map(optionalJobExecution -> optionalJobExecution
         .orElseThrow(() -> new NotFoundException(format("JobExecution with id '%s' was not found", jobExecutionId))))
-      .map(this::verifyJobExecution)
-      .map(this::modifyJobExecutionToCompleteWithCancelledStatus)
-      .compose(jobExec -> updateJobExecutionWithSnapshotStatus(jobExec, params))
-      .compose(jobExec -> deleteRecordsFromSRSIfNecessary(jobExec, params))
+      .compose(jobExecution -> {
+        try {
+          verifyJobExecution(jobExecution);
```
okolawole-ebsco, line 253: does `verifyJobExecution` return a future or any object?

VRohach: Hi @okolawole-ebsco ! It returns JobExecution object, without Future. I want to add that I will change this chain via "recover()"-method as @Abdulkhakimov said. Will refactor this.

VRohach: @okolawole-ebsco  I changed this approach via "recover"-method. Could you please re-review ?

Abdulkhakimov (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -496,8 +504,8 @@ private JobExecution verifyJobExecution(JobExecution jobExecution) {
     if (jobExecution.getStatus() == JobExecution.Status.ERROR || jobExecution.getStatus() == COMMITTED
     || jobExecution.getStatus() == JobExecution.Status.CANCELLED) {
       String msg = String.format("JobExecution with status '%s' cannot be forcibly completed", jobExecution.getStatus());
-      LOGGER.error(msg);
-      throw new BadRequestException(msg);
+      LOGGER.info(msg);
+      throw new JobUpdateException(msg);
```
SerhiiNosko, line 508: Maybe it makes sense to rename this exception something like this: JobUpdateDuplicateException (when we trying to update status that was already updated previously).
It will be more clear to answer question why we logging it as info, and why for it we return succeeded future

SerhiiNosko (approved):

PR #656, opened 2022-06-06
Reduce DI_COMPLETE, DI_ERROR consumers to 1

SerhiiNosko:
## Purpose
Right now we have 1 consumer of DI_RAW_RECORDS_CHUNK_READ, make the same for DI_COMPLETED/DI_ERROR consumer.

## Approach
Make DI_COMPLETE/DI_ERROR consumer's count the same as for DI_RAW_RECORDS_CHUNK_READ to avoid situation if DI_COMPLETE events processed by another instance


mod-source-record-manager-server/src/main/java/org/folio/rest/impl/InitAPIImpl.java
```diff
@@ -46,7 +46,7 @@ public class InitAPIImpl implements InitAPI {
   @Value("${srm.kafka.StoredMarcChunkConsumer.instancesNumber:3}")
   private int storedMarcChunkConsumerInstancesNumber;
 
-  @Value("${srm.kafka.DataImportConsumersVerticle.instancesNumber:3}")
+  @Value("${srm.kafka.DataImportConsumersVerticle.instancesNumber:1}")
```
VRohach, line 49: @SerhiiNosko does this change will not affect performance on production envs ?

RuslanLavrov (approved):

Abdulkhakimov (approved):

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/InitAPIImpl.java
SerhiiNosko: It should not affect performance in the negative side, because each consumer should process single partition. It does not make sense to increase number of consumers if we have the same number of partitions.
For scaling if we add more topic partitions - need to add more instances.

VRohach: Got it!Thank you.

VRohach (approved):

PR #660, opened 2022-06-10
MODSOURMAN-811. Ensure proper work of flow control in multi-instances and multi-partitions envs

SerhiiNosko:
## Purpose
Change in-memory counter to database counter for flow control feature to fix behaviour with multi-instances and multi-partition envs.

## Approach
Previously situations when 1 srm instance reads chunk and increases value of counter and 2 second instance consumes completed events and decreases own 2 srm instance counter were possible.
In order to fix it approach with database counter has been introduced(approach with hazelcast has not been considered due to complexity of integration and tuning).

During run of 50k Create import on perf rancher with 2 instances of srm, 2 partitions of DI_RAW_RECORDS_CHUNK_READ and DI_COMPLETED topics no any pauses and slowness observed with flow control with DB counter implementation.

**Perf results and comparition with previous releases:**
50K Create (Instances, Holdings, Items):
Juniper & Kiwi: 2h 30min+
Lotus: 1h 30min
MG: 1h

## Learning
https://issues.folio.org/browse/MODSOURMAN-811


mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDao.java
```diff
@@ -7,12 +7,49 @@
...
 
   /**
-   * Saves eventId and handlerId to database
+   * This method saves eventId and handlerId to deduplication table.
    *
    * @param handlerId handler id
    * @param eventId event id
    * @param tenantId tenant id
-   * @return future if the event not processed yet
+   * @return successful future if event has not been processed already, or failed future otherwise
```
VRohach, line 15: @SerhiiNosko Do you mean vice versa? successful if the event has been processed?

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/InitAPIImpl.java
```diff
@@ -43,13 +43,13 @@ public class InitAPIImpl implements InitAPI {
   @Value("${srm.kafka.RawMarcChunkConsumer.instancesNumber:1}")
   private int rawMarcChunkConsumerInstancesNumber;
 
-  @Value("${srm.kafka.StoredMarcChunkConsumer.instancesNumber:3}")
+  @Value("${srm.kafka.StoredMarcChunkConsumer.instancesNumber:1}")
```
VRohach, line 46: @SerhiiNosko  I think we need to mention these changes in release notes. What do you think?

mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -206,6 +206,16 @@
       "run": "after",
       "snippet": "CREATE INDEX IF NOT EXISTS journal_records_source_id_idx ON journal_records USING btree (source_id);",
       "fromModuleVersion": "mod-source-record-manager-3.4.0"
+    },
+    {
+      "run": "after",
```
VRohach, line 211: This is point for me... I need to re-check migration for srm more time after your changes will be merged to master.

VRohach (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDaoImpl.java
```diff
@@ -19,11 +19,18 @@ public class EventProcessedDaoImpl implements EventProcessedDao {
 
   private static final Logger LOGGER = LogManager.getLogger();
 
-  private static final String TABLE_NAME = "events_processed";
+  public static final String EVENTS_PROCESSED_TABLE_NAME = "events_processed";
+  public static final String FLOW_CONTROL_EVENTS_COUNTER_TABLE_NAME = "flow_control_events_counter";
 
   private static final String INSERT_SQL = "INSERT INTO %s.%s (handler_id, event_id) VALUES ($1, $2)";
 
-  private PostgresClientFactory pgClientFactory;
```
okolawole-ebsco, line 26: parameterized queries. ❤️

mod-source-record-manager-server/src/main/resources/templates/db_scripts/schema.json
SerhiiNosko: It would be good to re-check schema migration with this changes, thanks

mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDaoImpl.java
SerhiiNosko: Yes, as always for new code, thanks:)

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/InitAPIImpl.java
SerhiiNosko: Updating page with recommended params for deployment to mention about this, will post link after update's complete, thanks

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -65,89 +69,96 @@ public void resetState() {
...
   @Override
-  public void trackRecordCompleteEvent() {
+  public void trackRecordCompleteEvent(String tenantId, Integer actualCounterValue) {
     if (!enableFlowControl) {
       return;
     }
 
-    int current = currentState.decrementAndGet();
+    if (actualCounterValue < 0) {
+      LOGGER.info("Tenant: [{}]. Current value less that zero, because of single record imports or resetting state, back to zero...", tenantId);
```
okolawole-ebsco, line 128: "less than" instead of "less that"?

mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDaoImpl.java
```diff
@@ -19,11 +19,18 @@ public class EventProcessedDaoImpl implements EventProcessedDao {
...
+  private static final String INSERT_EVENT_AND_DECREASE_FLOW_CONTROL_COUNTER_SQL = "SELECT save_event_and_decrease_flow_control_counter($1, $2)";
+
+  private static final String DECREASE_COUNTER_SQL = "UPDATE %s.%s SET events_to_process = events_to_process - $1 RETURNING *";
```
okolawole-ebsco, line 29: thinking you should reference the specific column instead of "*". If additional column is added to the table, it will not break the receiving code in the module. I get that this is a generalized query that could be for multiple tables but i think the schema name should be substituted but the table name should be constant.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
SerhiiNosko: corrected, thanks

mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDaoImpl.java
SerhiiNosko: agree, changed code to return specific field

mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDao.java
SerhiiNosko: Yes, its correct, basically its deduplication table, if event already processed - it should return failed future with constraint violation exception, and our service recognized it and throws DuplicateEventException.
If event has not been yet processed - event id should be inserted in this table, and successful future will be returned

Abdulkhakimov (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDaoImpl.java
```diff
@@ -33,13 +40,60 @@ public EventProcessedDaoImpl(PostgresClientFactory pgClientFactory) {
...
+      pgClientFactory.createInstance(tenantId).execute(query, queryParams, promise);
+    } catch (Exception e) {
+      LOGGER.error("Failed to update counter value {} to table {}", counterValue, FLOW_CONTROL_EVENTS_COUNTER_TABLE_NAME, e);
+      promise.fail(e);
+    }
+    return getCounterValueFromRowSet(promise);
+  }
+
+  private Future<Integer> getCounterValueFromRowSet(Promise<RowSet<Row>> promise) {
+    return promise.future().map(resultSet -> resultSet.iterator().next().getInteger(0));
```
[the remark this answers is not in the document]
SerhiiNosko: Counter will not be updated, but in any case processing will be continued, and we have scheduled job that resets counter state to 0 each 5 min

mod-source-record-manager-server/src/main/java/org/folio/dao/EventProcessedDao.java
SerhiiNosko: Updated documentations https://wiki.folio.org/pages/viewpage.action?pageId=59932680. Here is what was changed:

- adjusted default values for consumer's count

image

- Add paragraph regarding topic's partition configuration
image

RuslanLavrov (approved):

PR #661, opened 2022-06-10
MODSOURMAN-810 Improve summary endpoint by parameter "entityType"

Abdulkhakimov:
## Purpose
Create ability to fetch job log entries filtering by entity type using new implemented param.

## TODOS and Open Questions

- [x] Add tests

## Learning
[MODSOURMAN-810](https://issues.folio.org/browse/MODSOURMAN-810)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
@@ -134,7 +134,7 @@ public class JournalRecordDaoImpl implements JournalRecordDao {
   private static final String SELECT_BY_JOB_EXECUTION_ID_QUERY = "SELECT * FROM %s.%s WHERE job_execution_id = $1";
   private static final String ORDER_BY_PATTERN = " ORDER BY %s %s";
   private static final String DELETE_BY_JOB_EXECUTION_ID_QUERY = "DELETE FROM %s.%s WHERE job_execution_id = $1";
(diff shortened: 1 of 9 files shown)


mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -134,7 +134,7 @@ public class JournalRecordDaoImpl implements JournalRecordDao {
   private static final String SELECT_BY_JOB_EXECUTION_ID_QUERY = "SELECT * FROM %s.%s WHERE job_execution_id = $1";
   private static final String ORDER_BY_PATTERN = " ORDER BY %s %s";
   private static final String DELETE_BY_JOB_EXECUTION_ID_QUERY = "DELETE FROM %s.%s WHERE job_execution_id = $1";
-  private static final String GET_JOB_LOG_ENTRIES_BY_JOB_EXECUTION_ID_QUERY = "SELECT * FROM get_job_log_entries('%s', '%s', '%s', %s, %s, %b)";
```
okolawole-ebsco, line 137: no need to change this for this PR due to time crunch. I would like to see more parameterized queries rather than string substitution approach that we currently use.

VRohach (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
Abdulkhakimov: I'll take a note

Abdulkhakimov:
> Is it possible to have a test for this?

Done

okolawole-ebsco (approved):

PR #663, opened 2022-06-15
MODSOURCE-509 Data Import Updates should add 035 field from 001/003, if it's not HRID or already exists. Removed HridService

AivarIUsupov:
## Purpose
When Inventory Single Record Import (or Data Import Update) is used to overlay an existing Instance, the resulting SRS MARC Bib should have correct 035 field (001 should not be HRID of an existing record or 035$a field already exists)

## Approach
Removed 001/003/035 fields manipulations from mod-source-record-manager and moved them to the mod-source-record-storage.

#### TODOS and Open Questions
- [ ] Merge only with mod-source-record-storage changes

## Learning
[MODSOURCE-509](https://issues.folio.org/browse/MODSOURCE-509)


okolawole-ebsco (approved):

RuslanLavrov (approved):

KaterynaSenchenko:
Thank you @Aliaksandr-Fedasiuk, merging this PR to unblock the other work on matching by 035s

PR #675, opened 2022-07-22
MODSOURMAN-823 - View all logs: broken alphabetical sorting via the "Job profile" column

RuslanLavrov:
## Purpose
there is spotted issue with sorting jobs by "Job profile" on the "view all" page. Rows with profile name in lower case placed after all rows having upper case in the profile name.
During investigation, it was found that issue is reproduced only on the snapshot, snapshot-2 envs. That might be due to another default collate type specified for database on the snapthot envs. On the vagrant env and perf-rancher env (and probably on mg-bugfest env where such issue is not reproduced as well), the "en_US.UTF-8" type is used.
image

## Approach
* convert to lower case `text` type columns to sort jobs by them case-insensitively
* add test

## Learning
[MODSOURMAN-823](https://issues.folio.org/browse/MODSOURMAN-823)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

VRohach:
@RuslanLavrov  Update NEWS please.

VRohach (approved):

PR #678, opened 2022-07-28
MODSOURCE-521 Populate 035 fields for exceptional cases of Update action

Aliaksandr-Fedasiuk:
Data Import Updates should add 035 field from 001/003, but only for exceptional Update scenarions (in such cases records are not prior saved in SRS, where 035 fields are handled)
035 fields should also be added if other 035 fields exist in the record

## Approach
Removing the creation of field 035 on the fly was erroneous because it caused an error: "MODSOURCE-521
Matching on created 035 does not work in Lotus or MG"
Remove the limitation on number of 035 fields
Construct 035 from 001 and 003 only for exceptional Update scenarios

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/HrIdFieldServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/HrIdFieldServiceImpl.java
@@ -1,0 +1,44 @@
+package org.folio.services.afterprocessing;
+
+import org.folio.rest.jaxrs.model.Record;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+import java.util.Objects;
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.folio.services.afterprocessing.AdditionalFieldsUtil.addDataFieldToMarcRecord;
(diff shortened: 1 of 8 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/HrIdFieldServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+  private static final String TAG_003 = "003";
+  private static final String TAG_035 = "035";
+  private static final char SUBFIELD_FOR_035 = 'a';
+  private static final char INDICATOR_FOR_035 = ' ';
+
+  @Override
+  public void move001valueTo035Field(List<Record> records) {
+    records.stream().parallel().forEach(record -> {
+      String valueFrom035 = getValue(record, TAG_035, SUBFIELD_FOR_035);
+      if (Objects.isNull(valueFrom035)) {
```
KaterynaSenchenko, line 27: So this line is the culprit for matching not working for records with multiple 035. Probably need to add the same logic as in SRS for adding 035 but not duplicating 035s.

VRohach (approved):

RuslanLavrov (approved):

PR #681, opened 2022-08-03
MODSOURMAN-840: Importing MARC records with 999 ff fields using Create jobs without match profiles causes data problems.

VRohach:
## Purpose
The main purpose is to Discard import via specific error message when there is 999ff$s or 999ff$i field in the record and action profile is Create Instance .

## Approach
There was added a mechanism which verify whether mrc-record contains 999ff$s or 999ff$i field and Create Instance action profile. In this case - there will be added an errorRecord with specific data in current record. It will handle in future flow.
Also, there was improved an error catching.

#### TODOS and Open Questions
The tests there should be fixed.

## Learning
(https://issues.folio.org/browse/MODSOURMAN-840)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -349,7 +349,7 @@ public class ChangeEngineServiceImpl implements ChangeEngineService {
   }
 
   private ParsedResult validateIf999ffFieldExistsOnInstanceCreateAction(JobExecution jobExecution, ParsedResult parsedResult) {
-    if (jobExecution.getJobProfileInfo().getDataType().equals(DataType.MARC)) {
+    if (jobExecution.getJobProfileInfo().getDataType().equals(DataType.MARC) && parsedResult.getParsedRecord()!= null) {
       var tmpRecord = new Record()
         .withParsedRecord(new ParsedRecord().withContent(parsedResult.getParsedRecord().encode()));
       if ((StringUtils.isNotBlank(getValue(tmpRecord, TAG_999, SUBFIELD_S)) && hasIndicator(tmpRecord, SUBFIELD_S))
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -101,6 +101,7 @@ public class ChangeEngineServiceImpl implements ChangeEngineService {
   private static final AtomicInteger indexer = new AtomicInteger();
   private static final String HOLDINGS_004_TAG_ERROR_MESSAGE =
     "The 004 tag of the Holdings doesn't has a link to the Bibliographic record";
+  public static final String INSTANCE_CANNOT_CREATED_999ff_ERROR_MESSAGE = "A new Instance was not created because the incoming record already contained a 999 ff";
 
   private final JobExecutionSourceChunkDao jobExecutionSourceChunkDao;
   private final JobExecutionService jobExecutionService;
@@ -239,6 +240,13 @@ public class ChangeEngineServiceImpl implements ChangeEngineService {
       Action.DELETE);
   }
 
+  private boolean createInstanceActionExists(JobExecution jobExecution) {
+    return containsMarcActionProfile(
+      jobExecution.getJobProfileSnapshotWrapper(),
+      List.of(FolioRecord.INSTANCE),
+      Action.CREATE);
+  }
+
   private boolean containsMarcActionProfile(ProfileSnapshotWrapper profileSnapshot,
                                                   List<FolioRecord> entityTypes, Action action) {
     List<ProfileSnapshotWrapper> childWrappers = profileSnapshot.getChildSnapshotWrappers();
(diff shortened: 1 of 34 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -338,6 +348,25 @@ public List<Record> getParsedRecordsFromInitialRecords(List<InitialRecord> rawRe
...
+      var tmpRecord = new Record()
+        .withParsedRecord(new ParsedRecord().withContent(parsedResult.getParsedRecord().encode()));
+      if ((StringUtils.isNotBlank(getValue(tmpRecord, TAG_999, SUBFIELD_S)) && hasIndicator(tmpRecord, SUBFIELD_S))
+        || (StringUtils.isNotBlank(getValue(tmpRecord, TAG_999, SUBFIELD_I)) && hasIndicator(tmpRecord, SUBFIELD_I))
+        && createInstanceActionExists(jobExecution)) {
+        ParsedResult result = new ParsedResult();
+        JsonObject errorObject = new JsonObject();
+        errorObject.put("error", INSTANCE_CREATION_999_ERROR_MESSAGE);
+        result.setErrors(errorObject);
+        result.setHasError(true);
```
okolawole-ebsco, line 362: This is not your problem. The hasError field should be automatically set when an errors object is set.

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/errorhandlers/payloadbuilders/MarcBibDiErrorPayloadBuilder.java
```diff
@@ -27,6 +27,7 @@
 @Component
 public class MarcBibDiErrorPayloadBuilder implements DiErrorPayloadBuilder {
   private static final String FIELDS = "fields";
+  public static final String NO_CONTENT_TITLE = "No content";
```
okolawole-ebsco, line 30: I think this string exists somewhere else. It would be good to unify all seperate instances into a util or constants class, so that the no content title can be uniform across the app.

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
VRohach: Hi @okolawole-ebsco! Agree. I changed method "isHasError" inside ParsedResult.

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
```diff
@@ -154,19 +154,22 @@ private JsonArray buildJournalRecords(List<Record> storedRecords, Optional<JsonO
...
-        .withEntityId(record.getId())
-        .withActionType(CREATE)
-        .withActionStatus(COMPLETED)
-        .withActionDate(new Date())
-        .withTitle(allNotNull(record.getParsedRecord(), titleFieldTag)
-          ? ParsedRecordUtil.retrieveDataByField(record.getParsedRecord(), titleFieldTag, subfieldCodes) : null);
-
-      journalRecords.add(JsonObject.mapFrom(journalRecord));
+
+      if (record.getErrorRecord() == null) {
```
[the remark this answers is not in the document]
VRohach: @okolawole-ebsco No. All needed logic will be triggered in the next data-import-flow steps. This condition is just required for NOT building journal record if there are some errors (because there can be NullPointerExceptions when trying to work with data from mrc-record)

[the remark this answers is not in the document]
okolawole-ebsco: understood.

okolawole-ebsco (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -338,6 +349,24 @@ public List<Record> getParsedRecordsFromInitialRecords(List<InitialRecord> rawRe
       }).collect(Collectors.toList());
   }
 
+  private ParsedResult validateIf999ffFieldExistsOnInstanceCreateAction(JobExecution jobExecution, ParsedResult parsedResult) {
```
Aliaksandr-Fedasiuk, line 352: @VRohach , I think it would be better to rename "validateIf999ffFieldExistsOnInstanceCreateAction" to "addErrorMessageWhen999ffFieldExists" because "validate" is better used for true/false results, and checking for CreateAction is not present in this method.

VRohach: @Aliaksandr-Fedasiuk  Good point! I will do it.

VRohach: @Aliaksandr-Fedasiuk  But there is a checking for CreateAction for Instance. 358 line: `&& createInstanceActionExists(jobExecution)`

VRohach: @Aliaksandr-Fedasiuk  I will rename on: "addErrorMessageWhen999ffFieldExistsOnInstanceCreateAction".

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

PR #682, opened 2022-08-07
MODSOURMAN-852: Fail job for unsupported profile

Aliaksandr-Fedasiuk:
## Purpose
DI stucks when using an unsupported profile

## Approach
Throw an UnsupportedProfileException and fail the job

## Learning
https://issues.folio.org/browse/MODSOURMAN-852


mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
```diff
@@ -56,6 +74,71 @@ public Future<Boolean> processChunk(RawRecordsDto incomingChunk, String jobExecu
...
+    for (ProfileSnapshotWrapper profileSnapshotWrapper : profileSnapshotWrapperList) {
+      if ((profileSnapshotWrapper.getContentType() == MATCH_PROFILE) && (isMatchingMarcBibToInstance(profileSnapshotWrapper))) {
+        ProfileSnapshotWrapper actionProfileSnapshotWrapper = getChildSnapshotWrapperByType(profileSnapshotWrapper, ACTION_PROFILE);
+        if (actionProfileSnapshotWrapper != null) {
+          ProfileSnapshotWrapper mappingProfileSnapshotWrapper = getChildSnapshotWrapperByType(actionProfileSnapshotWrapper, MAPPING_PROFILE);
+          if ((mappingProfileSnapshotWrapper != null) && (isMappingMarcBibToMarcBib(mappingProfileSnapshotWrapper))) {
+            return true;
+          }
+        }
+        isExistsMatchProfileToInstanceWithActionUpdateMarcBib(profileSnapshotWrapper.getChildSnapshotWrappers());
```
okolawole-ebsco, line 96: I am fuzzy on the business rules but thinking this line should have a `return` here; some result passed back to the base case of your recursive approach. As is, if the line returns true or false, it does not seem to have any impact. Please correct me if I am wrong.

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
```diff
@@ -56,6 +74,71 @@ public Future<Boolean> processChunk(RawRecordsDto incomingChunk, String jobExecu
...
+    return result;
+  }
+
+  private boolean isMappingMarcBibToMarcBib(ProfileSnapshotWrapper profileSnapshotWrapper) {
+    MappingProfile mappingProfile = extractProfile(profileSnapshotWrapper, MappingProfile.class);
+    return mappingProfile.getIncomingRecordType() == EntityType.MARC_BIBLIOGRAPHIC
+      && mappingProfile.getExistingRecordType() == EntityType.MARC_BIBLIOGRAPHIC;
+  }
+
+  private <T> T extractProfile(ProfileSnapshotWrapper profileSnapshotWrapper, Class clazz) {
```
okolawole-ebsco, line 132: Is there any exception handling should this method fail? class cast or json decode exceptions.

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
Aliaksandr-Fedasiuk: Yes. You are absolutely right. Thank you. I lost return during a lot of code reviews. 
Fixed.

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
Aliaksandr-Fedasiuk: This exception will be handled at a higher level.
Example:
image

okolawole-ebsco (approved):

mod-source-record-manager-server/src/test/java/org/folio/verticle/consumers/RawMarcChunkConsumersVerticleTest.java
```diff
@@ -41,28 +27,22 @@
...
+import static java.util.Collections.singletonList;
 import static org.apache.kafka.clients.consumer.ConsumerConfig.GROUP_ID_CONFIG;
 import static org.folio.dataimport.util.RestUtil.OKAPI_URL_HEADER;
-import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_ERROR;
-import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_MARC_FOR_UPDATE_RECEIVED;
-import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHUNK_PARSED;
-import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHUNK_READ;
-import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.ACTION_PROFILE;
-import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.JOB_PROFILE;
+import static org.folio.rest.jaxrs.model.DataImportEventTypes.*;
```
VRohach, line 39: @Aliaksandr-Fedasiuk Could you please remove "*"? Just set a huge value in imports in intellijidea.

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
```diff
@@ -56,6 +74,74 @@ public Future<Boolean> processChunk(RawRecordsDto incomingChunk, String jobExecu
...
+        if (snapshotWrapper.getContentType() == contentType) {
+          return snapshotWrapper;
+        }
+      }
+    }
+    return null;
+  }
+
+  private boolean isMatchingMarcBibToInstance(ProfileSnapshotWrapper profileSnapshotWrapper) {
+    boolean result = false;
```
VRohach, line 119: @Aliaksandr-Fedasiuk  I think it would be better to rename "isMatched" or "matched" to represent a more logical variable.

VRohach (approved):

RuslanLavrov (approved):

mod-source-record-manager-server/src/test/java/org/folio/verticle/consumers/RawMarcChunkConsumersVerticleTest.java
Aliaksandr-Fedasiuk: Ready.

mod-source-record-manager-server/src/main/java/org/folio/services/exceptions/UnsupportedProfileException.java
```diff
@@ -0,0 +1,7 @@
+package org.folio.services.exceptions;
+
+public class UnsupportedProfileException extends RuntimeException {
```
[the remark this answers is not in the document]
Aliaksandr-Fedasiuk: Ready.

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
Aliaksandr-Fedasiuk: Ready.

PR #683, opened 2022-08-13
MODSOURMAN-852: Clarification of condition for the 035 field creation

Aliaksandr-Fedasiuk:
## Purpose
SRM creates an extra 035 with hrId.

## Approach
Clarification of condition for the 035 field creation: when the MARC BIB record contains the 999i subfield this means that the current record exists in DB and we don't need to generate the 035 field because of a high probability the field 001 contains hrId.

## Learning
[MODSOURCE-528](https://issues.folio.org/browse/MODSOURCE-528)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk:
> 

Fixed.

okolawole-ebsco (approved):

RuslanLavrov (approved):

PR #684, opened 2022-08-17
MODSOURMAN-854: Importing MARC Authority and Holdings records with 999 marc fields, with default Create job profiles, causes data problems

VRohach:
## Purpose
This is the PR for extending already merged and tested functionality(https://issues.folio.org/browse/MODSOURMAN-840), but for "MARC_HOLDINGS" and "MARC_AUTHORITY" types. The primary purpose is to fail data-import if there are 999ff$s or 999ff$i in marc-files and create MARC_HOLDINGS or MARC_AUTHORITY entities in Action Profile.

## Approach
Added mechanism for these 2 new types, as it was added for the MARC_BIB. 

## Learning
JIRA: https://issues.folio.org/browse/MODSOURMAN-854


okolawole-ebsco (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -253,8 +255,22 @@ private boolean createInstanceActionExists(JobExecution jobExecution) {
       Action.CREATE);
   }
 
+  private boolean createAuthorityActionExists(JobExecution jobExecution) {
```
Aliaksandr-Fedasiuk, line 258: this method returns boolean, I think there should be better to name this method smth like "isCreateAuthorityActionExists". What do you think?

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -253,8 +255,22 @@ private boolean createInstanceActionExists(JobExecution jobExecution) {
...
+    return containsMarcActionProfile(
+      jobExecution.getJobProfileSnapshotWrapper(),
+      List.of(FolioRecord.AUTHORITY),
+      Action.CREATE);
+  }
+
+  private boolean createMarcHoldingsActionExists(JobExecution jobExecution) {
```
Aliaksandr-Fedasiuk, line 265: this method returns boolean, I think there should be better to name this method smth like "isCreateMarcHoldingsActionExists". What do you think?

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
VRohach: Agree. Thank you for this comment. I just did it as an analogy for already existing method "createInstanceActionExists". Yes, I will rename all of them.

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
VRohach: @Aliaksandr-Fedasiuk  Done.

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/errorhandlers/payloadbuilders/MarcAuthorityDiErrorPayloadBuilder.java
```diff
@@ -0,0 +1,45 @@
...
+                                                          Record currentRecord) {
+
+    return mappingRuleCache.get(new MappingRuleCacheKey(okapiParams.getTenantId(), currentRecord.getRecordType()))
+      .compose(rulesOptional -> {
+        DataImportEventPayload diErrorPayload = DiErrorBuilderUtil.prepareDiErrorEventPayload(throwable, okapiParams, jobExecutionId, currentRecord);
+        if (rulesOptional.isPresent()) {
+          return Future.succeededFuture(DiErrorBuilderUtil.makeLightweightPayload(currentRecord, NO_TITLE_MESSAGE, diErrorPayload));
+        }
+        return Future.succeededFuture(DiErrorBuilderUtil.makeLightweightPayload(currentRecord, null, diErrorPayload));
+      });
```
psmagin, line 43: As I can see rules are not used, so why it is there?

VRohach: @psmagin  good point! Removed it. Thank you. Could you please double-check and approve if everything is Ok ?

psmagin (approved):

PR #693, opened 2022-09-16
MODSOURMAN-874: Data import: fails the creation of a Holding through a match on the 999 ff field

VRohach:
## Purpose
To avoid invalid validation if MARC_BIB Holdings(not MARC-Holdings) wanna be created.

## Approach
Added additional condition to retrieve needed field from mapping profile.

## Learning
JIRA: https://issues.folio.org/browse/MODSOURMAN-874

### DIFF
--- a/mod-source-record-manager-server/src/test/java/org/folio/rest/impl/AbstractRestTest.java
+++ b/mod-source-record-manager-server/src/test/java/org/folio/rest/impl/AbstractRestTest.java
@@ -29,13 +29,11 @@ import io.vertx.ext.unit.TestContext;
 import net.mguenther.kafka.junit.EmbeddedKafkaCluster;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.header.internals.RecordHeader;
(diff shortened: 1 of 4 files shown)


Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/test/java/org/folio/rest/impl/AbstractRestTest.java
```diff
@@ -172,6 +172,12 @@ public abstract class AbstractRestTest {
     .withAction(ActionProfile.Action.CREATE)
     .withFolioRecord(ActionProfile.FolioRecord.HOLDINGS);
 
+/*  private final MappingProfile marcHoldingsMappingProfile = new MappingProfile()
```
okolawole-ebsco, line 175: is this intentional? the comment?

VRohach: @okolawole-ebsco  Good point. I will remove it.

RuslanLavrov (approved):

PR #703, opened 2022-10-13
[MODSOURMAN-887] Fixed sorting in the Status column

RomanChernetskyi:
## Purpose
Fixed sorting in the Status column

Jira: https://issues.folio.org/browse/MODSOURMAN-887


mod-source-record-manager-server/src/main/java/org/folio/rest/impl/MetadataProviderImpl.java
```diff
@@ -218,8 +218,18 @@ private List<SortField> mapSortQueryToSortFields(List<String> sortQuery) {
...
+      if (sortField.equals("status")) {
+        fields.add(new SortField("progress_current", getOppositeSortOrder(sortOrder)));
+      }
     }
     return fields;
   }
 
+  private String getOppositeSortOrder(String sortOrder) {
+    if (sortOrder.equals("asc")) {
+      return "desc";
```
Aliaksandr-Fedasiuk, line 231: @RomanChernetskyi, could we use string constants for "asc" and "desc" values?

Aliaksandr-Fedasiuk (approved):

KyryloShulzhenko (approved):

KaterynaSenchenko (approved):

PR #720, opened 2022-11-22
[MODSOURMAN-891]SRS MARC Created when No Create Action in Job Profile

KyryloShulzhenko:
https://issues.folio.org/browse/MODSOURMAN-891
Overview: When you want to create either a holdings connected to an existing instance or an item connected to an existing holdings, you match to either that holdings hrid or instance hrid and have an action to create either the holdings or item. What is happening is that the holdings and items are created AND the log also says the SRS MARC is created.

Current workaround: No workaround

Steps to Reproduce:

Log into FOLIO-snapshot as diku_admin
Go to Settings/Data Export
Create a new mapping
SRS and Holdings
Map Holdings HRID to 901$a
Map Holdings Id to 902$a
Create a new data export job linked to that mapping
Go to Settings/Data Import
Create a new field mapping for an item
Item mat type: Electronic resource
Loan type: Online
Status: Available
Create a new action "Create item" linked to that item field mapping
Create a new match for 901$a to Holdings hrid
Create a new job with that match and on matches, Creates item record
Go to Inventory
Find or create a Source = MARC Instance with 1 holdings record and no item records
Copy that Instance HRID and search for it in Inventory, so that you have only 1 search result
Go to Actions/Save instance UUIDs
Go to Data Export
Upload the UUIDs text file that was just downloaded
Apply the export job profile created in step 4 and export the file
Download the newly-created MARC file
Go to Data Import
Upload that exported mrc file and select test job to create item
Item is created
Log also says SRS Marc bibs are created
Case 2: For Holdings creation:

follow the same steps as above
except find an Inventory instance with no holdings
Create a field mapping to create a holdings record, and an associated action profile
create a field mapping for holdings record and a match from 001 to instance hrid.
Expected Results: The log should have a dash for all the MARC SRS bibs and instances. There's no action to create/update srs marc bibs or instances in either case

Actual Results: I suspect that this is creating orphan srs bibs whenever you create a holdings attached to an existing instance or an item attached to an existing holdings.

Additional Information: This is happening in snapshot and Morning Glory.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -159,7 +159,7 @@ public class ChangeEngineServiceImpl implements ChangeEngineService {
       .onSuccess(parsedRecords -> {
         fillParsedRecordsWithAdditionalFields(parsedRecords);
 
-        if (updateMarcActionExists(jobExecution) || updateInstanceActionExists(jobExecution)) {
+        if (updateMarcActionExists(jobExecution) || updateInstanceActionExists(jobExecution) || isCreateOrUpdateItemOrHoldingsActionExists(jobExecution, parsedRecords)) {
           hrIdFieldService.move001valueTo035Field(parsedRecords);
           updateRecords(parsedRecords, jobExecution, params)
             .onSuccess(ar -> promise.complete(parsedRecords))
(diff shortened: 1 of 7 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -267,6 +267,13 @@ private boolean updateInstanceActionExists(JobExecution jobExecution) {
       Action.UPDATE);
   }
 
+  private boolean isCreateItemActionExists(JobExecution jobExecution) {
```
KaterynaSenchenko, line 270: This condition should be wider - we need to skip the "save records in SRS" part for profiles that
**contain Create/Update actions on Holdings/Item, but no Create Instance**

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -267,6 +267,15 @@ private boolean updateInstanceActionExists(JobExecution jobExecution) {
       Action.UPDATE);
   }
 
+  private boolean isCreateOrUpdateItemOrHoldingsActionExists(JobExecution jobExecution, List<Record> parsedRecords) {
+    var jobProfileSnapshotWrapper = jobExecution.getJobProfileSnapshotWrapper();
+    var itemAndHoldingsList = List.of(FolioRecord.ITEM, FolioRecord.HOLDINGS);
+    return (containsMarcActionProfile(jobProfileSnapshotWrapper, itemAndHoldingsList, Action.CREATE) ||
+      containsMarcActionProfile(jobProfileSnapshotWrapper, itemAndHoldingsList, Action.UPDATE)) &&
+      !containsMarcActionProfile(jobProfileSnapshotWrapper, List.of(FolioRecord.INSTANCE), Action.CREATE) &&
+      parsedRecords.get(0).getRecordType() == MARC_BIB;
```
VRohach, line 276: @KyryloShulzhenko  Could you please double-check if there can be a situation when` parsedRecords.get(0)` will be null? If yes, let's add a check somewhere.

VRohach (commented):
Please update NEWS-file.

Aliaksandr-Fedasiuk (approved):

PR #730, opened 2022-12-27
[MODSOURMAN-927]No title displays in import log when SRS MARC and Ins…

KyryloShulzhenko:
When an import job contains create/update matches/actions for holdings and/or items, but no matches/actions for SRS MARC Bibs or Instances, no title is displayed in the log for the records in the imported file. Thus, if there is an error, it is difficult to identify which record had an error and what the error was (since there is no way to reach the record's JSON from the UI without the log title link)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
```diff
@@ -108,7 +107,8 @@ public void handle(JournalService journalService, DataImportEventPayload eventPa
 
   private Future<JournalRecord> populateRecordTitleIfNeeded(JournalRecord journalRecord,
                                                             DataImportEventPayload eventPayload) {
-    var entityType = journalRecord.getEntityType();
+    var entityType = (journalRecord.getEntityType() == HOLDINGS || journalRecord.getEntityType() == ITEM ?
+      MARC_BIBLIOGRAPHIC : journalRecord.getEntityType());
 
     if (entityType == MARC_BIBLIOGRAPHIC || entityType == MARC_AUTHORITY) {
```
KaterynaSenchenko, line 113: Could we just broaden this condition instead of setting the wrong entityType (MARC_BIBLIOGRAPHIC for Holdings and Item) above?

KyryloShulzhenko: Hello @KaterynaSenchenko, this function extract title from MARC_BIB. only in MARC_BIB has title indifferent cases such as MARC_BIB Holding or ITEM. It can be resolved 
else if (entityType == HOLDINGS || entityType == ITEM) {
      journalRecord.setTitle(NO_TITLE_MESSAGE);
      String recordAsString = eventPayload.getContext().get(MARC_BIBLIOGRAPHIC.value());
      if (StringUtils.isNotBlank(recordAsString)) {
        var parsedRecord = Json.decodeValue(recordAsString, Record.class).getParsedRecord();
        return mappingRuleCache.get(new MappingRuleCacheKey(eventPayload.getTenant(), MARC_BIBLIOGRAPHIC))
          .compose(ruleOptional -> ruleOptional
            .map(mappingRules -> {
              var titleExtractor = titleExtractorMap.get(entityType);
              if (titleExtractor == null || parsedRecord == null) {
                return null;
              }

              return titleExtractor.apply(parsedRecord, mappingRules);
            })
            .map(title -> Future.succeededFuture(journalRecord.withTitle(title)))
            .orElseGet(() -> Future.succeededFuture(journalRecord)));
      }
    }

RomanChernetskyi (approved):

PR #740, opened 2023-01-23
MODSOURMAN-937: Send DI_MARC_BIB_FOR_ORDER_CREATED event for Importing Orders

VRohach:
## Purpose
To differentiate the flow of importing Orders from regular DI, when MARC Bibs are saved in SRS a different event should be issued to be received by mod-orders. SRM should publish DI_MARC_BIB_FOR_ORDER_CREATED. Event type should be determined based on the action for creating Orders in the JobProfileWrSnapshotWrapper retrieved from JobExecution.

## Approach
Get JobExecution from DB and retrieve JobProfileSnapshotWrapper from it. If there is action on Order+Create, then set eventType for sending "DI_MARC_BIB_FOR_ORDER_CREATED"

## Learning
JIRA: https://issues.folio.org/browse/MODSOURMAN-937

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
@@ -120,6 +120,7 @@ public class StoredRecordChunksKafkaHandler implements AsyncRecordHandler<String
           return jobExecutionService.getJobExecutionById(jobExecutionId, okapiConnectionParams.getTenantId())
             .compose(jobExecutionOptional -> {
               if (jobExecutionOptional.isPresent()) {
+                LOGGER.debug("handle:: JobExecution found by id {}: chunkId:{} chunkNumber: {} ", jobExecutionId, chunkId, chunkNumber);
                 return setOrderEventTypeIfNeeded(jobExecutionOptional.get(), eventType);
               } else {
                 LOGGER.warn("handle:: Couldn't find JobExecution by id {}: chunkId:{} chunkNumber: {} ", jobExecutionId, chunkId, chunkNumber);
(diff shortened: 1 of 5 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
```diff
@@ -102,8 +117,17 @@ public Future<String> handle(KafkaConsumerRecord<String, String> record) {
 
           LOGGER.debug("handle:: RecordsBatchResponse has been received, starting processing chunkId: {} chunkNumber: {} jobExecutionId: {}", chunkId, chunkNumber, jobExecutionId);
           saveCreatedRecordsInfoToDataImportLog(storedRecords, okapiConnectionParams.getTenantId());
-          return recordsPublishingService.sendEventsWithRecords(storedRecords, jobExecutionId,
-              okapiConnectionParams, eventType.value())
+          return jobExecutionService.getJobExecutionById(jobExecutionId, okapiConnectionParams.getTenantId())
+            .compose(jobExecutionOptional -> {
+              if (jobExecutionOptional.isPresent()) {
```
Aliaksandr-Fedasiuk, line 122: @VRohach , could you add some debug information here and in other places with new changes?

Aliaksandr-Fedasiuk (approved):

RomanChernetskyi (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
VRohach: @Aliaksandr-Fedasiuk  Done.

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #747, opened 2023-01-31
[MODSOURMAN-932] Updated migrations scripts

RomanChernetskyi:
## Purpose
Prevent 500 error while schema upgrade

## Approach
Added DROP for get_job_execution_summary function and updated DROP for get_job_log_entries function


KaterynaSenchenko (approved):

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_job_execution_summary_function.sql
```diff
@@ -1,3 +1,6 @@
+-- Request to delete get_job_log_entries with a different return format.
```
RuslanLavrov, line 1:
```suggestion
-- Request to delete get_job_execution_summary with a different return format.
```

Aliaksandr-Fedasiuk (approved):

PR #749, opened 2023-02-03
MODSOURMAN-946: Handle DI_ERROR event for POLines

VRohach:
## Purpose
The main purpose is to handle DI_ERROR for Order/PO_Line importing for updating journal_record errors.

## Approach
There are two scenarios:
1. Update all records where entityType = PO_LINE and orderId = "{orderId}" and add current one with the same data.
2. Create only one record if  "orderId==null" and write the current record with the error message from the payload.

## Learning
JIRA: https://issues.folio.org/browse/MODSOURMAN-946

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
@@ -117,7 +117,6 @@ public class MarcImportEventsHandler implements SpecificEventHandler {
   }
 
   private void processJournalRecordForOrder(JournalService journalService, DataImportEventPayload eventPayload, String tenantId, JournalRecord journalRecord) {
-    populateOrderTitleIfNeeded(journalRecord, eventPayload);
     if (journalRecord.getOrderId() != null) {
       journalRecordService.updateErrorJournalRecordsByOrderIdAndJobExecution(journalRecord.getJobExecutionId(), journalRecord.getOrderId(), journalRecord.getError(), tenantId)
         .onComplete(e -> journalService.save(JsonObject.mapFrom(journalRecord), tenantId));
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
@@ -107,8 +107,8 @@ public class MarcImportEventsHandler implements SpecificEventHandler {
       JournalRecord journalRecord = JournalUtil.buildJournalRecordByEvent(eventPayload,
         journalParams.journalActionType, journalParams.journalEntityType, journalParams.journalActionStatus);
 
-      if (Objects.equals(journalRecord.getEntityType(), PO_LINE) && journalRecord.getError() != null) {
-        processJournalRecordForOrder(journalService, eventPayload, tenantId, journalRecord);
+      if (Objects.equals(journalRecord.getEntityType(), PO_LINE)) {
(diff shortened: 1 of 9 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
```diff
@@ -100,8 +107,32 @@ public void handle(JournalService journalService, DataImportEventPayload eventPa
       JournalRecord journalRecord = JournalUtil.buildJournalRecordByEvent(eventPayload,
         journalParams.journalActionType, journalParams.journalEntityType, journalParams.journalActionStatus);
 
-      populateRecordTitleIfNeeded(journalRecord, eventPayload)
-        .onComplete(ar -> journalService.save(JsonObject.mapFrom(journalRecord), tenantId));
+      if (Objects.equals(journalRecord.getEntityType(), PO_LINE) && journalRecord.getError() != null) {
```
okolawole-ebsco, line 110: if the journal record is a po line but getError is null, I was expecting `processJouirnalRecordForOrder` to process the record. 
If there is no error in the journal record, will `getError` return null?

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
```diff
@@ -100,11 +107,23 @@ public void handle(JournalService journalService, DataImportEventPayload eventPa
       JournalRecord journalRecord = JournalUtil.buildJournalRecordByEvent(eventPayload,
         journalParams.journalActionType, journalParams.journalEntityType, journalParams.journalActionStatus);
 
-      populateRecordTitleIfNeeded(journalRecord, eventPayload)
-        .onComplete(ar -> journalService.save(JsonObject.mapFrom(journalRecord), tenantId));
+      if (Objects.equals(journalRecord.getEntityType(), PO_LINE)) {
+        processJournalRecordForOrder(journalService, tenantId, journalRecord);
+      } else {
+        populateRecordTitleIfNeeded(journalRecord, eventPayload)
+          .onComplete(ar -> journalService.save(JsonObject.mapFrom(journalRecord), tenantId));
```
okolawole-ebsco, line 114: nitpick: `onComplete(ar -> journalService.save(JsonObject.mapFrom(journalRecord), tenantId))` is repeated here and twice in the `processJournalRecordForOrder` method. I think having  one representation for all code paths would be good.

okolawole-ebsco (approved):

RomanChernetskyi (approved):

KaterynaSenchenko (approved):

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
VRohach: Hello @okolawole-ebsco ! I investigated, and it seems like it would be better to add a check on `journalRecord.getError() != null` to the `processJournalRecordForOrder`-method inside ` if (journalRecord.getOrderId() != null) ` case. I think that it would a**void executing script** for the update if there is NO error (for NOT Failed) PO_LINEs. I added it to my PR. Do you have any objections regarding it?

okolawole-ebsco: The location of the check inside `processJournalRecordForOrder` makes sense to me. Looks good!

VRohach: @okolawole-ebsco thanks!

PR #753, opened 2023-02-10
[MODSOURMAN-939]Handle EDIFACT parsing exceptions - complete job with…

KyryloShulzhenko:
https://issues.folio.org/browse/MODSOURMAN-939
EDIFACT parsing errors cause import job to hang - handle such cases to complete job with errors.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/journal/InvoiceUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/journal/InvoiceUtil.java
@@ -11,9 +11,11 @@ import org.folio.DataImportEventPayload;
 import org.folio.rest.jaxrs.model.JournalRecord;
 import org.folio.rest.jaxrs.model.Record;
 
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/journal/InvoiceUtil.java
```diff
@@ -52,13 +56,16 @@ public class InvoiceUtil {
   private static final String NO_INVOICE_TITLE_MESSAGE = "No content";
   private static final String NO_VENDOR_INVOICE_NUMBER = "0";
 
-  public static LinkedList<JournalRecord> buildJournalRecordByEvent(DataImportEventPayload event)
+  public static List<JournalRecord> buildJournalRecordByEvent(DataImportEventPayload event)
```
okolawole-ebsco, line 59: Just for my knowledge, what is the thinking behind this change?

KyryloShulzhenko: I used List<JournalRecord> buildErrorJournalRecord() which returns "List" at 67 line and I think It`s improved code by SOLID principles

RuslanLavrov (approved):

PR #783, opened 2023-05-23
MODSOURMAN-990 Add migration mechanism for mapping rules

psmagin:
## Purpose
Add migration mechanism for mapping rules

## Approach
Create Migration service for custom code-based migrations

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/migration/impl/HoldingsMapping852CallNumberTypeCustomMigration.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/migration/impl/HoldingsMapping852CallNumberTypeCustomMigration.java
@@ -1,0 +1,61 @@
+package org.folio.services.migration.impl;
+
+import static org.folio.Record.RecordType.MARC_HOLDING;
+
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonArray;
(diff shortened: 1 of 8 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/migration/impl/HoldingsMapping852CallNumberTypeCustomMigration.java
```diff
@@ -0,0 +1,61 @@
...
+    if (tag852Rules == null) {
+      return rules;
+    }
+    for (int i = 0; i < tag852Rules.size(); i++) {
+      var entities = tag852Rules.getJsonObject(i).getJsonArray("entity");
+      if (entities != null) {
+        for (int j = 0; j < entities.size(); j++) {
+          var rule = entities.getJsonObject(j);
+          if ("callNumberTypeId".equals(rule.getString("target"))) {
+            rule.put("subfield", new JsonArray(List.of("b")));
```
Shans-Kaluhin, line 54: You can use this one
```suggestion
            rule.put("subfield", JsonArray.of("b"));
```

Shans-Kaluhin (approved):

viacheslavkol (approved):

PR #784, opened 2023-05-25
[MODSOURCEMAN-993]Create journal_records for non matched entities during multiple matching

KyryloShulzhenko:
## Links
https://issues.folio.org/browse/MODSOURMAN-993
## Purpose
Add multiple Holdings/Items and Multiple Non Matches handle journal records support

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
@@ -1,6 +1,8 @@
 package org.folio.services.journal;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import org.apache.commons.lang3.StringUtils;
 import org.folio.DataImportEventPayload;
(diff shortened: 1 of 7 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/JournalParams.java
```diff
@@ -221,6 +229,14 @@ public Optional<JournalParams> getJournalParams(DataImportEventPayload eventPayl
           JournalRecord.ActionStatus.COMPLETED));
       }
     },
+    DI_INVENTORY_ITEM_MATCHED {
+      @Override
+      public Optional<JournalParams> getJournalParams(DataImportEventPayload eventPayload) {
+        return Optional.of(new JournalParams(JournalRecord.ActionType.NON_MATCH,
+          JournalRecord.EntityType.HOLDINGS,
```
RuslanLavrov, line 236:
```suggestion
    DI_INVENTORY_ITEM_MATCHED {
      @Override
      public Optional<JournalParams> getJournalParams(DataImportEventPayload eventPayload) {
        return Optional.of(new JournalParams(JournalRecord.ActionType.MATCH,
          JournalRecord.EntityType.ITEM,
```

mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
```diff
@@ -59,49 +64,82 @@ record = new Record()
...
-          if (entityType == PO_LINE){
-            journalRecord.setOrderId(entityJson.getString("purchaseOrderId"));
-          }
-          journalRecord.setEntityHrId(entityJson.getString("hrid"));
+      if(isNotBlank(eventPayloadContext.get("NOT_MATCHED_NUMBER"))) {
+        int notMatchedNumber = Integer.parseInt(eventPayloadContext.get("NOT_MATCHED_NUMBER"));
+        for (int i = 0; i<notMatchedNumber;i++) {
+          journalRecords.add(constructBlankJournalRecord(record, JournalRecord.ActionType.NON_MATCH,entityType,actionStatus)
+            .withEntityType(entityType)
+            .withEntityId(UUID.randomUUID().toString()));
```
RuslanLavrov, line 74: propose not to set a random uuid as entityId to journalRecord when holdings/items are not found

PR #785, opened 2023-06-05
Modsourman 989

Aliaksandr-Fedasiuk:
## Purpose
Previous Flow Control implementation heavily loaded the DB for module synchronization. 

## Approach
Using a fetch(long) operator for managing to receive packages from DI_RAW_RECORDS_CHUNK_READ topic.

[MODSOURMAN-989](https://issues.folio.org/browse/MODSOURMAN-989)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
@@ -12,6 +12,7 @@ import org.springframework.stereotype.Service;
 
 import javax.annotation.PostConstruct;
 import java.util.Map;
+import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHUNK_READ;
@@ -21,7 +22,7 @@ import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHU
 public class RawRecordsFlowControlServiceImpl implements RawRecordsFlowControlService {
   private static final Logger LOGGER = LogManager.getLogger();
 
-  private static final Integer instanceId = (int)(Math.random() * 10 + 1);
+  private static final Integer instanceId = new Random().nextInt(9);
 
   @Value("${di.flow.control.max.simultaneous.chunks:2}")
   private Integer maxSimultaneousChunks;
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
@@ -137,8 +137,12 @@ public class RawRecordsFlowControlServiceImpl implements RawRecordsFlowControlSe
         LOGGER.info("resumeIfThresholdAllows :: Tenant: [{}]. ConsumerId:{}, Demand:{}, Current state:{}, History state: {}",
           tenantId, consumer.getId(), consumer.demand(), currentState.get(tenantId), historyState.get(tenantId));
         if (((consumer.demand() == 0) && (currentState.get(tenantId) < recordsThreshold)) || historyState.get(tenantId).getValue() > 0) {
-          consumer.fetch(maxSimultaneousRecords);
 
+          if (consumer.demand() > maxSimultaneousRecords) {
+            consumer.pause();
+          }
+
+          consumer.fetch(maxSimultaneousRecords);
           LOGGER.info("resumeIfThresholdAllows :: Tenant: [{}]. Fetch :: ConsumerId: {}, Demand:{}, History state: {}",
             tenantId, consumer.getId(), consumer.demand(), historyState.get(tenantId));
 
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
@@ -131,7 +131,8 @@ public class RawRecordsFlowControlServiceImpl implements RawRecordsFlowControlSe
         LOGGER.info("resumeIfThresholdAllows :: Tenant: [{}]. DI_RAW_RECORDS_CHUNK_READ. " +
             "ConsumerId:{}, Demand:{}, Current state:{}, History state: {}",
           tenantId, consumer.getId(), consumer.demand(), currentState.get(tenantId), historyState.get(tenantId));
-        if (((consumer.demand() == 0) && (currentState.get(tenantId) < recordsThreshold)) || historyState.get(tenantId).getValue() > 0) {
+        if (((consumer.demand() == 0) && (currentState.get(tenantId) < recordsThreshold)) ||
+          (historyState.get(tenantId).getKey() > 0 && historyState.get(tenantId).getValue() > 0)) {
           if (consumer.demand() > maxSimultaneousChunks) {
             consumer.pause(); //set demand to 0, because fetch can only add new demand value when demand > 0
           }
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
@@ -45,16 +45,6 @@ public class RawRecordsFlowControlServiceImpl implements RawRecordsFlowControlSe
   @PostConstruct
   public void init() {
     LOGGER.info("init:: Flow control feature is {}", enableFlowControl ? "enabled" : "disabled");
-
-    // A read stream is either in "flowing" or "fetch" mode
-    // initially the stream is in "flowing" mode
-    // when the stream is in "flowing" mode, elements are delivered to the handler
-    // when the stream is in "fetch" mode, only the number of requested elements will be delivered to the handler
-    consumersStorage.getConsumersByEvent(DI_RAW_RECORDS_CHUNK_READ.value())
-      .forEach(consumer -> {
-        consumer.pause();
-        consumer.fetch(maxSimultaneousChunks);
-      });
   }
 
   /**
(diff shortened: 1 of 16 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -22,61 +21,61 @@
...
+   * This method schedules resetting state of flow control. By default, it triggered each 2 mins.
    * This is distributed system and count of raw records pushed to process can be not corresponding
    * with DI_COMPLETE/DI_ERROR, for example because of imports are stuck.
    *
-   * For example, batch size - 10, max simultaneous - 20, records threshold - 10.
-   * Flow control receives 20 records and pause consumers, 5 DI_COMPLETE events came, 15 was missed due to some DB issue.
-   * Current state would equals to 15 and threshold will never met 10 records so consumers will be paused forever
+   * For example, batch size - 50 records, max simultaneous - 1 batch, records threshold - 25.
+   * Flow control receives 50 records (1 batch) and pause consumers, 20 DI_COMPLETE events came, 30 was missed due to some DB issue.
+   * Current state would equals to 30 and threshold will never met 24 records so consumers will be paused forever
```
okolawole-ebsco, line 62: `threshold will never met 24 records` should this be 25 instead since your records threshold is set to 25?

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -85,80 +84,99 @@ public void trackChunkReceivedEvent(String tenantId, Integer recordsCount) {
...
-          consumer.resume();
+    LOGGER.info("resumeIfThresholdAllows :: Tenant: [{}]. Current state:{}, History state: {}",
+      tenantId, currentState.get(tenantId), historyState.get(tenantId));
 
-          LOGGER.info("resumeIfThresholdAllows:: Tenant: [{}]. Kafka consumer - id: {}, subscription - {} is resumed for all tenants, because {} met threshold {}",
-            tenantId, consumer.getId(), DI_RAW_RECORDS_CHUNK_READ.value(), this.currentState, recordsThreshold);
+    historyState.putIfAbsent(tenantId, Pair.of(0,0));
+    if (historyState.get(tenantId).getKey().equals(currentState.get(tenantId))) {
+      Integer previousValue = historyState.get(tenantId).getValue();
+      historyState.put(tenantId, Pair.of(historyState.get(tenantId).getKey(), ++previousValue));
```
okolawole-ebsco, line 129: I think you need to use `computeIfPresent` instead of getting a previous value and then incrementing it. This action should happen atomically, getting and incrementing by one.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -85,80 +84,99 @@ public void trackChunkReceivedEvent(String tenantId, Integer recordsCount) {
...
+      historyState.put(tenantId, Pair.of(currentState.get(tenantId),0));
+    }
+
+    consumersStorage.getConsumersByEvent(DI_RAW_RECORDS_CHUNK_READ.value())
+      .forEach(consumer -> {
+        LOGGER.info("resumeIfThresholdAllows :: Tenant: [{}]. DI_RAW_RECORDS_CHUNK_READ. " +
+            "ConsumerId:{}, Demand:{}, Current state:{}, History state: {}",
+          tenantId, consumer.getId(), consumer.demand(), currentState.get(tenantId), historyState.get(tenantId));
+        if (((consumer.demand() == 0) && (currentState.get(tenantId) < recordsThreshold)) ||
+          (historyState.get(tenantId).getKey() > 0 && historyState.get(tenantId).getValue() > 0)) {
```
okolawole-ebsco, line 142: it think you should get the key and value from historyState at the same time so that your comparison can be consistent.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -85,80 +84,99 @@ public void trackChunkReceivedEvent(String tenantId, Integer recordsCount) {
...
         }
       });
+  }
+
+  private void initFetchMode(String tenantId) {
+
+    // A read stream is either in "flowing" or "fetch" mode
+    // initially the stream is in "flowing" mode
+    // when the stream is in "flowing" mode, elements are delivered to the handler
+    // when the stream is in "fetch" mode, only the number of requested elements will be delivered to the handler
```
okolawole-ebsco, line 164: on the handler is done processing a fetch, will another fetch be performed automatically?

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
```diff
@@ -12,6 +12,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHUNK_READ;
```
okolawole-ebsco, line 15: Looks like this might not be used in this file.

mod-source-record-manager-server/src/main/resources/log4j2.properties
```diff
@@ -18,23 +18,23 @@ logger.folio_kafka.level = debug
 logger.folio_kafka.additivity = false
 logger.folio_kafka.appenderRef.stdout.ref = STDOUT
 
-rootLogger.level = info
-rootLogger.appenderRefs = info
+rootLogger.level = debug
```
okolawole-ebsco, line 21: I am concerned that is going against recent logging improvements and not cluttering log files. What is your thinking behind this?

mod-source-record-manager-server/src/main/resources/vertx-default-jul-logging.properties
```diff
@@ -1,5 +1,5 @@
 handlers = java.util.logging.ConsoleHandler
 .level = ALL
 
-logger.cql2pgjson.level = ERROR
-logger.cql2pgjson.name = org.folio.cql2pgjson.CQL2PgJSON
+#logger.cql2pgjson.level = ERROR
```
okolawole-ebsco, line 4: is this intentional?

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
Aliaksandr-Fedasiuk: Not. Fetch could not be started automatically. For this, I use a scheduler for checking a consumer and when a consumer is on pause (demand == 0), the scheduler will increase demand.

mod-source-record-manager-server/src/main/java/org/folio/verticle/AbstractConsumersVerticle.java
Aliaksandr-Fedasiuk: Thank you. Fixed.

mod-source-record-manager-server/src/main/resources/log4j2.properties
Aliaksandr-Fedasiuk: Changes in logs related to the testing process. I reverted these changes. Thank you.

mod-source-record-manager-server/src/main/resources/vertx-default-jul-logging.properties
Aliaksandr-Fedasiuk: Changes in logs related to the testing process. I reverted these changes. Thank you.

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
Aliaksandr-Fedasiuk: Ready. Thanks

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
Aliaksandr-Fedasiuk: Fixed. Thanks

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
```diff
@@ -49,31 +53,31 @@ public class RawRecordsFlowControlServiceImpl implements RawRecordsFlowControlSe
...
-  @Scheduled(initialDelayString = "PT1M", fixedRateString = "${di.flow.control.reset.state.interval:PT2M}")
+  @Scheduled(initialDelayString = "PT1M", fixedRateString = "${di.flow.control.reset.state.interval:PT1M}")
   public void resetState() {
     if (!enableFlowControl) {
       return;
     }
 
     currentState.forEach((tenantId, counterVal) -> {
-      LOGGER.info("resetState:: Tenant: [{}] ", tenantId);
+      LOGGER.info("resetState:: Tenant: [{}], InstanceIs:{} ", tenantId, instanceId);
```
okolawole-ebsco, line 80: should the log be `InstanceId:` instead of `InstanceIs:`

other locations have this potential issue as well.

Aliaksandr-Fedasiuk: oh, thanks. I added a unique id because now in logs I can't separate log messages from different SRM instances

mod-source-record-manager-server/src/main/java/org/folio/services/flowcontrol/RawRecordsFlowControlServiceImpl.java
okolawole-ebsco: Reached out to @Aliaksandr-Fedasiuk on Slack. The scheduler is not responsible for calling fetch on the consumer each time. It is responsible for waking a paused consumer. The consumer will continue to fetch messages on a cadence.

okolawole-ebsco (approved):

KaterynaSenchenko (approved):

PR #792, opened 2023-07-18
MULTIPLE IMPORT

RomanChernetskyi:
## Purpose
Implement logic for multiple import at srm

## Approach
[[MODSOURMAN-965]](https://github.com/folio-org/mod-source-record-manager/pull/773) Adjust progress tracking mechanism with regard to multiple values being created
[[MODSOURMAN-971]](https://github.com/folio-org/mod-source-record-manager/pull/770) Adjust journal records population to create multiple journal records for each Holdings/Item
[[MODSOURMAN-975]](https://github.com/folio-org/mod-source-record-manager/pull/780) Adjust DI log DTO to allow displaying multiple Holding/Item
[[MODSOURMAN-993](https://github.com/folio-org/mod-source-record-manager/pull/784)] Create journal_records for non matched entities during multiple matching


Aliaksandr-Fedasiuk (approved):

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -360,31 +365,44 @@ private RecordProcessingLogDto mapRowSetToRecordProcessingLogDto(RowSet<Row> res
...
-            .withError(row.getString(PO_LINE_ENTITY_ERROR))
-            .withOrderId(row.getString(ORDER_ENTITY_ID)))
-        .withRelatedInvoiceInfo(constructProcessedEntityInfoBasedOnEntityType(row,
-          INVOICE_ACTION_STATUS, INVOICE_ENTITY_ID, INVOICE_ENTITY_HRID, INVOICE_ENTITY_ERROR))
-        .withRelatedInvoiceLineInfo(constructInvoiceLineInfo(row)));
+    List<ProcessedHoldingsInfo> processedHoldingsInfo = new LinkedList<>();
+    List<ProcessedItemInfo> processedItemInfo = new LinkedList<>();
+
+    Row row = resultSet.iterator().next();
+    recordProcessingLogSummary
```
yaroslav-epam, line 372: `recordProcessingLogSummary`  creation could be moved under the `resultSet` size check thus it would not be created in case of empty `resultSet`

mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
```diff
@@ -70,38 +83,152 @@ record = new ObjectMapper().readValue(recordAsString, Record.class);
...
-        if (entityType == INSTANCE || entityType == HOLDINGS || entityType == ITEM || entityType == PO_LINE) {
+      if ((actionType == JournalRecord.ActionType.MATCH || actionType == JournalRecord.ActionType.NON_MATCH)
+        && (entityType == HOLDINGS || entityType == ITEM)) {
+        List<JournalRecord> resultedJournalRecords = new ArrayList<>();
+        if (isNotBlank(eventPayloadContext.get(NOT_MATCHED_NUMBER))) {
+          int notMatchedNumber = Integer.parseInt(eventPayloadContext.get(NOT_MATCHED_NUMBER));
+          for (int i = 0; i < notMatchedNumber; i++) {
+            resultedJournalRecords.add(constructBlankJournalRecord(record, JournalRecord.ActionType.NON_MATCH, entityType, actionStatus)
+              .withEntityType(entityType)
+              .withEntityId(UUID.randomUUID().toString()));
```
RuslanLavrov, line 98: propose not to set a random uuid as entityId to journalRecord when holdings/items are not found

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
RomanChernetskyi: Thanks, done

mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
RomanChernetskyi: Thanks, done

PR #794, opened 2023-07-30
MODSOURMAN-1011: Import An Instance With A Known Identifier

Aliaksandr-Fedasiuk:
## Purpose
Instances need to be created by Data Import(DI) with a known identifier to realize ECS. For local instance that have source=MARC, DI will create the Shared Instance in the central tenant. The Shared instance needs to be created with the same identifier as the local instance that is being promoted

## Approach
This is not currently allowed in DI. A query parameter will be added to signal to DI to allow the 999ff$i field to be set and also create an instance with value in 999ff$i.
Query parameter will be accepted when initializing a data import at /change-manager/jobExecutions/<id>/records
Query parameter will be called acceptInstanceId and settable to true/false with a default of false.

## Links
[MODSOURMAN-1011](https://issues.folio.org/browse/MODSOURMAN-1011)

### DIFF
--- a/mod-source-record-manager-server/src/test/java/org/folio/services/ChangeEngineServiceImplTest.java
+++ b/mod-source-record-manager-server/src/test/java/org/folio/services/ChangeEngineServiceImplTest.java
@@ -1,10 +1,55 @@
 package org.folio.services;
 
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonObject;
+import io.vertx.kafka.client.producer.KafkaHeader;
+import org.folio.MatchProfile;
+import org.folio.dao.JobExecutionSourceChunkDao;
+import org.folio.dataimport.util.OkapiConnectionParams;
+import org.folio.dataimport.util.marc.MarcRecordAnalyzer;
+import org.folio.dataimport.util.marc.MarcRecordType;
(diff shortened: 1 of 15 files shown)


RomanChernetskyi (approved):

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/test/java/org/folio/services/ChangeEngineServiceImplTest.java
```diff
@@ -36,9 +36,8 @@
 import io.vertx.kafka.client.producer.KafkaHeader;
 
 import org.folio.MatchProfile;
-import org.folio.rest.jaxrs.model.EntityType;
-import org.folio.rest.jaxrs.model.MappingMetadataDto;
-import org.folio.rest.jaxrs.model.MappingProfile;
+import org.folio.rest.jaxrs.model.*;
```
yaroslav-epam, line 39: Should we use the wildcard import?

yaroslav-epam (approved):

mod-source-record-manager-server/src/test/java/org/folio/services/ChangeEngineServiceImplTest.java
Aliaksandr-Fedasiuk: completed, thanks

RuslanLavrov (approved):

PR #800, opened 2023-09-05
MODSOURMAN-1031: Add support for a single holding

yaroslav-epam:
## Purpose
_Add support for a single holding which comes from mod-inventory while journaling._

## Approach
_Currently mod-inventory sends a single holding as JsonObject so SRM should handle both cases single and multiple_

## Tested
image


RomanChernetskyi (approved):

VRohach (approved):

mod-source-record-manager-server/src/test/java/org/folio/services/JournalUtilTest.java
```diff
@@ -163,6 +163,39 @@ public void shouldBuildJournalRecordWhenNoEntityPresent() throws JournalRecordMa
...
+    String snapshotId = UUID.randomUUID().toString();
+
+    JsonObject recordJson = new JsonObject()
+      .put("id", recordId)
+      .put("snapshotId", snapshotId)
+      .put("order", 1);
+
+    HashMap<String, String> context = new HashMap<>();
+    context.put(HOLDINGS.value(), holdingsJson.encode());
+    context.put(MARC_BIBLIOGRAPHIC.value(), recordJson.encode());
```
RuslanLavrov, line 187: propose to use MARC_HOLDINGS instead of MARC_BIBLIOGRAPHIC to describe this case of marc-holdings import more accurately

RuslanLavrov (approved):

PR #804, opened 2023-09-13
MODDATAIMP-911 add the totalRecordsInFile field to the Entity and DTO

TarasSpashchenko:
## Purpose
MODDATAIMP-911 add the totalRecordsInFile field to the Entity and DTO to support correct progress bar in UI-data-import application.


ncovercash (approved):
lgtm. what branch of raml-storage should I use in DI?

PR #805, opened 2023-09-18
[MODSOURMAN-1041] Check for composite completion upon external status updates

ncovercash:
# [Jira MODSOURMAN-1041](https://issues.folio.org/browse/MODSOURMAN-1041)

## Purpose
When a composite job's child completes, we may need to mark its parent as completed, too.  This was done as part of the normal processing flow [here](https://github.com/folio-org/mod-source-record-manager/blob/6f135cae9f91c1a773631c134385d0182990a816/mod-source-record-manager-server/src/main/java/org/folio/services/RecordProcessedEventHandlingServiceImpl.java#L124), however, we should also do this upon external status changes from the `status` endpoint.

## Approach
This adds similar code as what is in `RecordProcessedEventHandlingServiceImpl`: when the update status method is called, we will first check that the updated job is a composite child; if so, we'll check that its siblings have all completed; and if all those conditions apply, we'll mark the parent `COMMITTED`/`RUNNING_COMPLETE`.

## Coverage: 💯 

image

## Extra
- Cleaned up some commented-out logs left in from previous DITF work
- Added a convenience method for tests to easily create parent and children composite job executions

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
@@ -384,8 +384,6 @@ public class JobExecutionServiceImpl implements JobExecutionService {
 
   private RunBy buildRunByFromUserInfo(UserInfo info) {
     RunBy result = new RunBy();
-//    LOGGER.warn("BAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
-//    LOGGER.warn("Got UserInfo {}", info);
     if (info != null) {
       result.setFirstName(info.getFirstName());
       result.setLastName(info.getLastName());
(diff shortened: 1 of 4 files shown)


TarasSpashchenko (approved):

okolawole-ebsco:
Any tests possible?

ncovercash:
@okolawole-ebsco added a set of tests for 100% coverage of these changes, including single/multiple-child composite jobs and applicable negative cases.

@TarasSpashchenko, can you confirm if this should be commented out or not? I am not sure; I know we were debugging it as part of the "run by" errors
image

okolawole-ebsco (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
```diff
@@ -190,7 +197,40 @@ public Future<JobExecution> updateJobExecutionStatus(String jobExecutionId, Stat
...
+            return this.getJobExecutionById(jobExecution.getParentJobId(), params.getTenantId())
+              .map(v -> v.orElseThrow(() -> new IllegalStateException("Could not find parent job execution")))
+              .compose(parentExecution -> 
+                this.getJobExecutionCollectionByParentId(parentExecution.getId(), 0, Integer.MAX_VALUE, params.getTenantId())
+                  .map(JobExecutionDtoCollection::getJobExecutions)
+                  // ensure all other children are completed
+                  .map(children ->
+                    children.stream()
+                      .filter(child -> child.getSubordinationType().equals(JobExecutionDto.SubordinationType.COMPOSITE_CHILD))
+                      .map(JobExecutionDto::getUiStatus)
```
RuslanLavrov, line 215: could you say please just for information, why the "uiStatus" field was chosen for checking the condition? Just the  "status" field seems more related to the backend.

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
ncovercash: Either would have been acceptable; the UI status seemed to be a little more future proof in my mind since even if additional processing/parsing/etc states are added in the feature, they'd show "completed" in the UI.  Also, since the problem was that we wanted these parent jobs to show as completed in the UI when there children were, I figured the UI's view was the best basis for this

ncovercash:
> @okolawole-ebsco added a set of tests for 100% coverage of these changes, including single/multiple-child composite jobs and applicable negative cases.
> 
> @TarasSpashchenko, can you confirm if this should be commented out or not? I am not sure; I know we were debugging it as part of the "run by" errors image

Looks like it's working without; we can probably remove this during later cleanup then. Sorry for the alarm; saw Sonar picked it up and wanted to make sure we hadn't accidentally removed significant code.

mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java
RuslanLavrov: yes, understand/agree that both fields are acceptable because, for every complete "status" field value, there is a corresponding "uiStatus" field value. Merely asked this because the "status" field is mostly used as the origin of a job status information in the module. For example as here in the method a bit above https://github.com/folio-org/mod-source-record-manager/blob/ae5cb38335d87c79e42fa9bb46186bccfa769f21/mod-source-record-manager-server/src/main/java/org/folio/services/JobExecutionServiceImpl.java#L189
Thanks for the explanation!

PR #807, opened 2023-09-26
MODSOURMAN-1045: allow create action when non-matches for instance

yaroslav-epam:
## Purpose
Allow create action when non-matches for instance without match profile

## Approach
In cases when there is no a match profile `reactTo` is `null` therefore it should be handled properly to construct `ParsedResult` with errors for such a case.

## Tested on Rancher
image

[MODSOURMAN-1045](https://issues.folio.org/browse/MODSOURMAN-1045)

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -11,7 +11,7 @@ import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_MARC_FOR_DELETE
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_MARC_FOR_UPDATE_RECEIVED;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_RAW_RECORDS_CHUNK_PARSED;
 import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.ACTION_PROFILE;
-import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ReactTo.MATCH;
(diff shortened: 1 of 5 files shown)


RomanChernetskyi (approved):

VRohach (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -293,15 +293,15 @@ private boolean deleteMarcActionExists(JobExecution jobExecution) {
   }
 
   private boolean isCreateInstanceActionExists(JobExecution jobExecution) {
-    return containsCreateInstanceActionWithMatch(jobExecution.getJobProfileSnapshotWrapper());
+    return containsCreateInstanceActionWithNonMatch(jobExecution.getJobProfileSnapshotWrapper());
   }
 
-  private boolean containsCreateInstanceActionWithMatch(ProfileSnapshotWrapper profileSnapshot) {
+  private boolean containsCreateInstanceActionWithNonMatch(ProfileSnapshotWrapper profileSnapshot) {
```
RuslanLavrov, line 299: propose to leave the previous method name (containsCreateInstanceActionWithMatch) because "containsCreateInstanceActionWithNonMatch" says about action profil with NON-MAR "reactTo" value but actually the method will return `false` if childWrapper.getReactTo() = NON-MATCH, and will return `true` 
if childWrapper.getReactTo() = MATCH or childWrapper.getReactTo() = null

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
yaroslav-epam: right, also was considering about that

RuslanLavrov: thanks!

PR #808, opened 2023-09-28
MODSOURMAN-1014: Upgrade folio-kafka-wrapper to 3.0.0 version

Maksat-Galymzhan:
## Purpose
Upgrade folio-kafka-wrapper to 3.0.0 version
## Approach
Upgrade the version of folio-kafka-wrapper to 3.0.0. Check out the readme at https://github.com/folio-org/folio-kafka-wrapper/blob/master/README.md for usage.

Acceptance Criteria
- kafka records are built using KafkaProducerRecordBuilder
- topic names are generated using KafkaTopicNameHelper

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
@@ -12,12 +12,15 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.kafka.KafkaConfig;
 import org.folio.kafka.KafkaTopicNameHelper;
+import org.folio.kafka.SimpleKafkaProducerManager;
+import org.folio.kafka.services.KafkaProducerRecordBuilder;
 import org.folio.processing.events.utils.PomReaderUtil;
 import org.folio.rest.jaxrs.model.Event;
 import org.folio.rest.jaxrs.model.EventMetadata;
 import org.folio.rest.jaxrs.model.RecordCollection;
 import org.folio.rest.tools.utils.ModuleName;
 import org.folio.services.exceptions.RecordsPublishingException;
+
 import static org.folio.services.util.RecordConversionUtil.RECORDS;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -116,8 +125,7 @@ public static String constructModuleName() {
   }
 
   public static KafkaProducer<String, String> createProducer(String eventType, KafkaConfig kafkaConfig) {
-    String producerName = eventType + "_Producer";
-    return KafkaProducer.createShared(Vertx.currentContext().owner(), producerName, kafkaConfig.getProducerProps());
+    return new SimpleKafkaProducerManager(Vertx.currentContext().owner(), kafkaConfig).createShared(eventType);
```
okolawole-ebsco, line 128: I think SimpleKafkaProducerManager is supposed to be a singleton.

mod-source-record-manager-server/src/main/java/org/folio/services/QuickMarcEventProducerServiceImpl.java
```diff
@@ -52,16 +52,16 @@ private Future<Boolean> sendEventInternal(String eventPayload, String eventType,
...
-        producer.write(record)
-          .onSuccess(unused -> {
+        producer.send(producerRecord)
+          .onSuccess(res -> {
             LOGGER.info("sendEventInternal:: Event with type {} was sent to kafka", eventType);
             promise.complete(true);
           })
-          .onFailure(throwable -> {
-            var cause = throwable.getCause();
+          .onFailure(err -> {
```
RomanChernetskyi, line 63: Could you please explain, what is the point of this refactoring?

RomanChernetskyi (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/QuickMarcEventProducerServiceImpl.java
Maksat-Galymzhan: perhaps, I shouldn't have touched it, since it just tautology, I roll it back to previous version

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
Maksat-Galymzhan: Correct if am wrong, but is it good idea to do singleton in util class?

VRohach (approved):

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
okolawole-ebsco: I see your point. since a new kafka config is one of the inputs to this method.

PR #813, opened 2023-10-19
MODSOURMAN-1064: Default log configuration is way too verbose

Aliaksandr-Fedasiuk:
## Purpose
Default log configuration is way too verbose

## Approach
Returned INFO level except DEBUG

[MODSOURMAN-1064](https://issues.folio.org/browse/MODSOURMAN-1064)


VRohach (approved):
@Aliaksandr-Fedasiuk Update NEWS pls.

PR #814, opened 2023-10-26
MODSOURMAN-1020: save incoming records for DI logs

yaroslav-epam:
## Purpose
Save DI logs as IncomingRecords in a new table

**Related PR**: https://github.com/folio-org/data-import-raml-storage/pull/300

## Approach
Save to DB IncomingRecords with JournalRecords after incoming records are parsed
 
## Testing
Debugged locally


KaterynaSenchenko:
@yaroslav-epam, please create a dev branch (with feature name for example) and open this PR on it. We won't merge it to master before all the pieces of functionality are ready

PR #816, opened 2023-10-31
MODSOURMAN-1021: Add endpoint for IncomingRecord 

yaroslav-epam:
## Purpose
Provide IncomingRecord API

## Approach
Add endpoint to get IncomingRecord by ID with DAO and service layers functionality

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/IncomingRecordDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/IncomingRecordDaoImpl.java
@@ -14,6 +14,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Repository;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 import static java.lang.String.format;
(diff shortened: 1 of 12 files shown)


KaterynaSenchenko (approved):

KaterynaSenchenko (approved):

Maksat-Galymzhan (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/IncomingRecordDaoImpl.java
```diff
@@ -24,14 +26,30 @@ public class IncomingRecordDaoImpl implements IncomingRecordDao {
...
   public static final String INCOMING_RECORDS_TABLE = "incoming_records";
+  private static final String GET_BY_ID_SQL = "SELECT * FROM %s.%s WHERE id = $1";
   private static final String INSERT_SQL = "INSERT INTO %s.%s (id, job_execution_id, incoming_record) VALUES ($1, $2, $3)";
 
   @Autowired
   private PostgresClientFactory pgClientFactory;
 
+  @Override
+  public Future<Optional<IncomingRecordDto>> getIncomingRecordDtoById(String id, String tenantId) {
+    LOGGER.debug("getById:: Get IncomingRecordDto by id = {} from the {} table", id, INCOMING_RECORDS_TABLE);
```
tatsiana-tarhonskaya, line 37: looks like you meant  getIncomingRecordDtoById instead of getById in both messages

tatsiana-tarhonskaya (approved):

ramls/metadata-provider.raml
```diff
@@ -242,4 +243,17 @@ resourceTypes:
...
             text/plain:
               example: "Internal server error"
+  /incomingRecords/{recordId}:
+    get:
+      description: get incoming record DTO by id
+      responses:
+        200:
+          body:
+            application/json:
+              type: incomingRecordDto
```
RuslanLavrov, line 253: propose to return incomingRecord itself instead of returning it as String value for this dto

yaroslav-epam: good point. discussed offline, adhering to this approach

RuslanLavrov (approved):

PR #835, opened 2023-12-19
MODSOURMAN-1108 MARC authority record is not created when use Job profile with match profile and action only on no-match branch

JavokhirAbdullayev:
## Purpose
MARC authority record is not created when use Job profile with match profile and action only on no-match branch. 
[MODSOURMAN-1108](https://issues.folio.org/browse/MODSOURMAN-1108)

## Approach
* replace condition for checking update action with match profile exist
* add test case
* fix tests

## Checklist
- [x] I have updated NEWS.md.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -4,7 +4,6 @@ import static java.lang.Boolean.TRUE;
 import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
-
 import static org.folio.rest.RestVerticle.MODULE_SPECIFIC_ARGS;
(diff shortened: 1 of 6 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -167,7 +165,7 @@ public Future<List<Record>> parseRawRecordsChunkForJobExecution(RawRecordsDto ch
       .onSuccess(parsedRecords -> {
         fillParsedRecordsWithAdditionalFields(parsedRecords);
 
-        if (updateMarcActionExists(jobExecution) || updateInstanceActionExists(jobExecution) || isCreateOrUpdateItemOrHoldingsActionExists(jobExecution, parsedRecords)) {
```
KaterynaSenchenko, line 170: We cannot delete these conditions, especially for Poppy - it will break many other flows. You could probably just add another OR clause with the check for MARC Authority profile

KaterynaSenchenko (approved):

RuslanLavrov (approved):

PR #836, opened 2023-12-26
Fixed handling of response deserialization error

RuslanLavrov:
## Purpose
to prevent import job hanging when response schema mismatches

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/MappingParametersProvider.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/MappingParametersProvider.java
@@ -19,6 +19,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/MappingParametersProvider.java
okolawole-ebsco, line 1: since the same pattern of changes occur over many lines, it makes sense to abstract the pattern into a method and call the method instead.

KaterynaSenchenko (approved):

okolawole-ebsco (approved):
Any tests possible?

mod-source-record-manager-server/src/main/java/org/folio/services/mappers/processor/MappingParametersProvider.java
RuslanLavrov: refactored

PR #841, opened 2024-01-11
MODSOURMAN-1113 Reduce Memory Allocations Of Strings

okolawole-ebsco:
## Purpose
Reduce memory allocations of strings to reduce memory requirements of SRS and improve performance & reliability at the limit

## Approach
Here are the several approaches involved:
- Changed the the cache in AdditionalFieldsUtil to enable stats only when metrics is enabled
- Utilized existing ObjectMappers instead of creating new ObjectMapper for each new JSON operation.
- In MappingMetadataService, instead of caching POJOs and then serializing them upon every HTTP request to get mapping metadata, the serialized POJO (as a string) is cached instead. This saves creating new String representations with each HTTP request, the strings can be large.
- In some handlers like one for journals, `currentNode` in `DataImportEventPayload` is ignored since it is not used and is expensive to create.
- Kafka consumers convert from byte[] to POJO. This included allowing KafkaConsumerWrappers to be able to configure value deserializer class.
- During journal processing, `DataImportEventPayload` is serialized into a String in a `Event` object. When `Event` is deserialized, it creates a large string for `DataImportEventPayload`. This large string is then deserialized into a `DataImportEventPayload` object. For example, if the paylaod was 1MB, the following object will be around 1MB
  - The byte array from Kafka consumer record
  - The `Event` POJO containing the payload string
  - The `DataImportEventPayload` POJO

  Deserialization has been changed such that intermediate String in the `Event` POJO is not created. This can be quite large.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/util/JournalEventDeserializer.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/util/JournalEventDeserializer.java
@@ -1,0 +1,56 @@
+package org.folio.util;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import io.vertx.core.json.jackson.DatabindCodec;
+
+import java.io.CharArrayReader;
+import java.io.IOException;
+
+/**
+ * Custom deserializer for {@link JournalEvent}. This deserializer understands that the
+ * {@link JournalEvent#eventPayload} is a string that has a JSON object embedded within.
+ * {@link JournalEvent#eventPayload} is deserialized without creating some intermediate String
+ * objects if this custom deserializer was not used.
+ * <p>
+ *   This was needed because the event payload is very large, containing "current node" information.
+ *   Multiple string representations of the event payload costed more memory than needed.
+ * </p>
+ */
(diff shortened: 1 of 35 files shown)


KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/util/JournalEventDeserializer.java
```diff
@@ -0,0 +1,56 @@
...
+    throws IOException {
+    JournalEvent journalEvent = new JournalEvent();
+
+    while (!jsonParser.isClosed()) {
+      JsonToken jsonToken = jsonParser.nextToken();
+
+      if (JsonToken.FIELD_NAME.equals(jsonToken)) {
+        String fieldName = jsonParser.getCurrentName();
+        jsonParser.nextToken(); // move to the value of the field
+        switch (fieldName) {
```
yaroslav-epam, line 35: we could use Java 17 switch statement

yaroslav-epam (approved):

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/util/JournalEventDeserializer.java
okolawole-ebsco: Hey! What advantage would a Java 17 switch provide?

PR #846, opened 2024-01-23
MODDATAIMP-957: add Orders creation

yaroslav-epam:
## Purpose
Fix stuck of Orders creation

## Approach
Add Order creation handling when create Order action received

## Is this change testable? If not - why?
Tested locally

## Checklist
- [ ] I have updated NEWS.md.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
@@ -8,6 +8,7 @@ import static org.folio.rest.RestVerticle.MODULE_SPECIFIC_ARGS;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_ERROR;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_INCOMING_EDIFACT_RECORD_PARSED;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_INCOMING_MARC_BIB_RECORD_PARSED;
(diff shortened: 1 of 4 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/ChangeEngineServiceImpl.java
```diff
@@ -196,6 +201,8 @@ private void processRecords(List<Record> parsedRecords, JobExecution jobExecutio
       }
       case DELETE_RECORD -> deleteRecords(parsedRecords, jobExecution, params)
         .onSuccess(ar -> promise.complete(parsedRecords)).onFailure(promise::fail);
+      case CREATE_ORDER -> sendEvents(parsedRecords, jobExecution, params, DI_MARC_BIB_FOR_ORDER_CREATED)
```
KaterynaSenchenko, line 204: should we consider renaming this event since we no longer create SRS MARC Bib for Orders?

yaroslav-epam: how about `DI_INCOMING_MARC_BIB_ORDER_PARSED`?

KaterynaSenchenko: DI_INCOMING_MARC_BIB_FOR_ORDER_PARSED

KaterynaSenchenko (approved):

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #847, opened 2024-01-25
MODSOURMAN-1106: Fixed error counter and blank title record

Maksat-Galymzhan:
## Purpose
MODSOURMAN-1106: Fixed error counter and blank title record

## Approach
Added default label if label is blank
Fixed error counter issue
Added sufficient amount of test cases related to job execution summary

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
@@ -43,7 +43,6 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static java.lang.String.format;
-import static org.apache.commons.lang3.ObjectUtils.allNotNull;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_INCOMING_EDIFACT_RECORD_PARSED;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_MARC_BIB_FOR_ORDER_CREATED;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_SRS_MARC_AUTHORITY_RECORD_CREATED;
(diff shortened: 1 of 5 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
```diff
@@ -226,8 +230,8 @@ private JsonArray buildJournalRecords(List<Record> storedRecords, Optional<JsonO
           .withActionType(CREATE)
           .withActionStatus(COMPLETED)
           .withActionDate(new Date())
-          .withTitle(allNotNull(record.getParsedRecord(), titleFieldTag)
-            ? ParsedRecordUtil.retrieveDataByField(record.getParsedRecord(), titleFieldTag, subfieldCodes) : null);
+          .withTitle(allNotNull(record.getParsedRecord(), titleFieldTag) ?
```
KaterynaSenchenko, line 233: didn't you make this check on line 222?

Maksat-Galymzhan: > didn't you make this check on line 222?

thanks for noticing it, fixed it

JavokhirAbdullayev (approved):

Aliaksandr-Fedasiuk (approved):

VRohach (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
```diff
@@ -164,7 +164,12 @@ private Future<JournalRecord> populateRecordTitleIfNeeded(JournalRecord journalR
 
               return titleExtractor.apply(parsedRecord, mappingRules);
             })
-            .map(title -> Future.succeededFuture(journalRecord.withTitle(title)))
+            .map(title -> {
+              if (title.isEmpty() || title == null) {
```
yaroslav-epam, line 168: please check first `title` is null then is empty

PR #853, opened 2024-01-30
[MODSOURMAN-1125] Fix error during authority creation

RomanChernetskyi:
## Purpose
Fix error during authority creation
## Approach
Prevent NPE if fildNAmeTag for title is null

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandler.java
@@ -180,8 +180,8 @@ public class StoredRecordChunksKafkaHandler implements AsyncRecordHandler<String
     for (Record record : storedRecords) {
 
       if (record.getErrorRecord() == null) {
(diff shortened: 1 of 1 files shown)


Maksat-Galymzhan (approved):

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/test/java/org/folio/verticle/consumers/StoredRecordChunksKafkaHandlerTest.java
```diff
@@ -138,6 +139,45 @@ public void shouldWriteSavedEdifactRecordsInfoToImportJournal() throws IOExcepti
     writeSavedRecordsInfoToImportJournal(EDIFACT_RECORD_PATH, EntityType.EDIFACT);
   }
 
+  @Test
+  public void shouldWriteSavedMarcBibRecordsInfoToImportJournalWhenTitleFieldTagIsNull() throws IOException {
```
RuslanLavrov, line 143:
```suggestion
  public void shouldWriteSavedMarcAuthorityRecordsInfoToImportJournalWhenTitleFieldTagIsNull() throws IOException {
```

VRohach (approved):
@RomanChernetskyi  Can you please update NEWS?

PR #854, opened 2024-02-05
FAT-9178 return job log entries for purchase order lines

okolawole-ebsco:
## Purpose
Return job log entries for purchase orders

## Approach
include the PO_LINE entity type in the list of expected entity types

### DIFF
--- a/mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
+++ b/mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
@@ -66,7 +66,7 @@ BEGIN
             null AS invoice_line_entity_error
       FROM
     	    (SELECT temp_result.job_execution_id, entity_id, temp_result.title, temp_result.source_record_order, action_type, error, temp_result.source_id, temp_result.tenant_id
-    	    FROM temp_result WHERE entity_type IN ('MARC_BIBLIOGRAPHIC', 'MARC_HOLDINGS', 'MARC_AUTHORITY')) AS marc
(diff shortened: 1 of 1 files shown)


Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_get_record_processing_log_function.sql
```diff
@@ -66,7 +66,7 @@ BEGIN
             null AS invoice_line_entity_error
       FROM
     	    (SELECT temp_result.job_execution_id, entity_id, temp_result.title, temp_result.source_record_order, action_type, error, temp_result.source_id, temp_result.tenant_id
-    	    FROM temp_result WHERE entity_type IN ('MARC_BIBLIOGRAPHIC', 'MARC_HOLDINGS', 'MARC_AUTHORITY')) AS marc
+    	    FROM temp_result WHERE entity_type IN ('MARC_BIBLIOGRAPHIC', 'MARC_HOLDINGS', 'MARC_AUTHORITY', 'PO_LINE')) AS marc
```
KaterynaSenchenko, line 69: Hi @okolawole-ebsco, could you please explain why this was added here? PO_LINE info is being formed at line 80, in this place we are dealing with MARCs

okolawole-ebsco: This subquery is the main table with which the subquery in line 80 is doing a left join. So if PO_LINE(s) are not present in this main subquery, it won't show up at all. The previous version of the function had the object separate and it was clearer in that instance. the subquery called `marc` is not really true.

Karate tests passed after this change. It may be worthwhile to revisit the whole function as these database object are too complex to understand and debug.

PR #858, opened 2024-02-19
refactor(reorder fields): keep order of MARC fields while Creating/Deriving/Editing MARC records. 

TsaghikKhachatryan:
## Purpose
Keep the order of MARC fields while Creating/Deriving/Editing MARC records. 

## Approach
Created a method reorderMarcRecordFields to rearrange fields after the content has been processed by the Marc Reader. This Reader often changes the order back, which prevents the desired outcome.

## Learning
https://folio-org.atlassian.net/browse/MODQM-407


mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -160,6 +171,59 @@ public static boolean addFieldToMarcRecord(Record record, String field, char sub
...
+      }
+
+      jsonNodesByTag.values().forEach(rearrangedArray::addAll);
+
+      ((ObjectNode)parsedContent).set("fields", rearrangedArray);
+
+      return parsedContent.toString();
+    } catch (Exception e) {
+      LOGGER.error("An error occurred while reordering Marc record fields: {}", e.getMessage(), e);
+      return null;
```
KaterynaSenchenko, line 199: If reodering failed, maybe it is better to log and return the sourceContent. If you return null for the parsedContent it will fail the further processing of record

KaterynaSenchenko (approved):

psmagin (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -50,6 +59,8 @@ public final class AdditionalFieldsUtil {
 
   private final static CacheLoader<Object, org.marc4j.marc.Record> parsedRecordContentCacheLoader;
   private final static LoadingCache<Object, org.marc4j.marc.Record> parsedRecordContentCache;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
```
okolawole-ebsco, line 62: For the future, consider reusing existing ObjectMappers.

PR #861, opened 2024-02-26
MODSOURMAN-1131 The import of file for creating orders is completed with errors

JavokhirAbdullayev:
## Purpose
The import of file for creating orders is completed with errors [MODSOURMAN-1131](https://folio-org.atlassian.net/browse/MODSOURMAN-1131)

## Approach
parse title from po line, update news

## Checklist
- [x] I have updated NEWS.md.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
@@ -175,12 +175,14 @@ public class MarcImportEventsHandler implements SpecificEventHandler {
     } else if (entityType == PO_LINE) {
       String recordAsString = eventPayload.getContext().get(entityType.value());
       if (StringUtils.isNotBlank(recordAsString)) {
-        var title = Optional.of(Json.decodeValue(recordAsString, RelatedPoLineInfo.class)
+        var title = Json.decodeValue(recordAsString, RelatedPoLineInfo.class)
           .getAdditionalProperties()
-          .get(PO_LINE_TITLE))
-          .map(Object::toString)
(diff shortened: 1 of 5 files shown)


dmytrokrutii (approved):

VRohach (approved):

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/consumers/util/MarcImportEventsHandler.java
```diff
@@ -172,8 +172,19 @@ private Future<JournalRecord> populateRecordTitleIfNeeded(JournalRecord journalR
             })
             .orElseGet(() -> Future.succeededFuture(journalRecord)));
       }
+    } else if (entityType == PO_LINE) {
+      String recordAsString = eventPayload.getContext().get(entityType.value());
+      if (StringUtils.isNotBlank(recordAsString)) {
+        var title = Json.decodeValue(recordAsString, RelatedPoLineInfo.class)
```
RuslanLavrov, line 178: propose to create JsonObject from PO_LINE value and use it to extract PO_LINE_TITLE field because RelatedPoLineInfo is not expected to be placed into the eventPayload. Instead, the poLine entity is expected to be there under PO_LINE key, but there is no corresponding class in the module. Propose not to try to deserialize RelatedPoLineInfo from the PO_LINE value (despite deserialization doesn't fail) to avoid wrong interpretation about having RelatedPoLineInfo in the payload during the code reading/maintaining.

JavokhirAbdullayev: thanks for suggestion, will re-open pr, after merging this one, because it contains code to fix kafka test issues on jenkins

PR #867, opened 2024-03-06
[MODSOURMAN-1141] Refactor source_id to be set during the build of journal records

RomanChernetskyi:
## Purpose
Refactor source_id to be set during the build of journal records

## Approach
Refactor source_id to be set during the build of journal records

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
@@ -55,6 +55,7 @@ public class JournalUtil {
   private static final String CENTRAL_TENANT_ID_KEY = "CENTRAL_TENANT_ID";
   private static final String CURRENT_EVENT_TYPE = "CURRENT_EVENT_TYPE";
   public static final String MARC_BIB_RECORD_CREATED = "MARC_BIB_RECORD_CREATED";
+  public static final String INCOMING_RECORD_ID = "INCOMING_RECORD_ID";
 
   private JournalUtil() {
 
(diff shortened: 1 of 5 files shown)


KaterynaSenchenko (approved):

RuslanLavrov (approved):

Maksat-Galymzhan (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/journal/JournalUtil.java
```diff
@@ -201,18 +203,18 @@ private static JournalRecord buildJournalRecordWithMarcBibType(JournalRecord.Act
...
 
   private static JournalRecord buildCommonJournalRecord(JournalRecord.ActionStatus actionStatus, JournalRecord.ActionType actionType, Record currentRecord,
-                                                        DataImportEventPayload eventPayload, HashMap<String, String> eventPayloadContext) {
+                                                        DataImportEventPayload eventPayload, HashMap<String, String> eventPayloadContext, String incomingRecordId) {
     String tenantId = eventPayload.getContext().get(CENTRAL_TENANT_ID_KEY);
 
     var currentJournalRecord = new JournalRecord()
       .withJobExecutionId(currentRecord.getSnapshotId())
-      .withSourceId(currentRecord.getId())
+      .withSourceId(incomingRecordId)
```
RuslanLavrov, line 217: looks like we need also to update the "get_job_log_entries" and "get_record_processing_log" functions so that they will return recordId from the "entity_id" column instead of "source_id" since now "source_id" column is being populated with incomingRecordId. Otherwise, suppose, ui will not get a record from srs to display on the log page.

PR #873, opened 2024-03-14
MODSOURMAN-956: Stop processing the job with incorrect profile

Maksat-Galymzhan:
## Purpose
Stop processing the job with incorrect profile
## Approach
Added logic to stop following chunks when job profile is not compatible with file
## Learning
[MODSOURMAN-956](https://folio-org.atlassian.net/browse/MODSOURMAN-956)

image

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/InvalidJobProfileForFileException.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/InvalidJobProfileForFileException.java
@@ -1,0 +1,12 @@
+package org.folio.services;
+
+public class InvalidJobProfileForFileException extends Exception {
+
+  public InvalidJobProfileForFileException(String message) {
+    super(message);
+  }
+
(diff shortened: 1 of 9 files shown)


Aliaksandr-Fedasiuk (approved):

dmytrokrutii (approved):

RuslanLavrov (approved):

JavokhirAbdullayev (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/InvalidJobProfileForFileException.java
```diff
@@ -0,0 +1,12 @@
+package org.folio.services;
+
+public class InvalidJobProfileForFileException extends Exception {
```
VRohach, line 3: @Maksat-Galymzhan  Please add documentation for this Exception.

VRohach (approved):

RomanChernetskyi (approved):

PR #895, opened 2024-04-19
[MODSOURMAN-1134] Filter MATCH journal_records by record_id during union

RomanChernetskyi:
## Purpose
Did not show discarded for MARC_AUTHORITY with error if other AUTHORITY updated at job
## Approach
Filter MATCH journal_records by record_id during union

### DIFF
--- a/mod-source-record-manager-server/src/test/java/org/folio/rest/impl/metadataProvider/MetaDataProviderJobLogEntriesAPITest.java
+++ b/mod-source-record-manager-server/src/test/java/org/folio/rest/impl/metadataProvider/MetaDataProviderJobLogEntriesAPITest.java
@@ -1090,6 +1090,45 @@ public class MetaDataProviderJobLogEntriesAPITest extends AbstractRestTest {
   }
 
   @Test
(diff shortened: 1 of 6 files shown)


mod-source-record-manager-server/src/test/java/org/folio/rest/impl/metadataProvider/MetaDataProviderJobLogEntriesAPITest.java
```diff
@@ -1089,6 +1089,45 @@ public void shouldReturnDiscardedForMarcBibAndInstanceIfMarcBibNotMatched(TestCo
     }));
   }
 
+  @Test
+  public void shouldReturnDiscardedForAuthorityIfAuthorityMatchedAndErrorAndOtherAuthorityUpdated(TestContext context) {
```
VRohach, line 1093: @RomanChernetskyi  Maybe it would be better to rename "shouldReturnDiscardedForAuthorityIfAuthorityMatched**With**ErrorAndOtherAuthorityUpdated" ?

KaterynaSenchenko (approved):

JavokhirAbdullayev (approved):

Aliaksandr-Fedasiuk (approved):

PR #897, opened 2024-04-22
MODSOURMAN-1127 LCCN normalization for MARC authority: change mappings for 010 field

viacheslavpoliakov:
## Purpose
MODSOURMAN-1127 Separate 010$a and 010$z subfield mapping as "LCCN" for $a and "Canceled LCCN" for $z

## Approach
Change mapping rule
Create migration class

## Checklist
- [x] I have updated NEWS.md.
- [x] I have added javadocs to new methods.


mod-source-record-manager-server/src/main/java/org/folio/services/migration/impl/AuthorityMapping010LccnCustomMigration.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private JsonObject updateRules(JsonObject rules) {
+    var tag010Rules = rules.getJsonArray("010");
+    if (tag010Rules == null) {
+      return rules;
+    }
+    for (int i = 0; i < tag010Rules.size(); i++) {
+      var entities = tag010Rules.getJsonObject(i).getJsonArray("entity");
+      if (entities != null) {
+        tag010Rules.getJsonObject(i).put("entity", FieldMappingHelper.get010FieldEntityJsonArray());
```
viacheslavkol, line 54: Can't we avoid constructing those rules from scratch if we just duplicate existing and remove unwanted subfields?

mod-source-record-manager-server/src/main/java/org/folio/services/migration/helper/FieldMappingHelper.java
psmagin, line 1: All of this is very specific to concrete migration, so I think it should be part of migration class.

viacheslavpoliakov: I decided to create separate helper file for mapping process in case of potential future mapping changes

viacheslavpoliakov: I've moved this logic to the migration class

psmagin (approved):

viacheslavkol (approved):

PR #903, opened 2024-05-08
MODSOURMAN-1186: "Something went wrong" error message after trying to view all logs in Data Import

Aliaksandr-Fedasiuk:
## Purpose
The result of the "Request file to be assembled for import" operation 
POST: /data-import/uploadDefinitions/<id>/files/<id>/assembleStorageFile 
returns "Internal Server Error, Please contact System Administrator or try again". 

Since jobProfileInfo appears only at the stage of import start, and import cannot start because of 
ERROR ileProcessingService Unable to initialize parent job: java.lang.NullPointerException: Cannot invoke "String.startsWith(String)" because "key" is null 
the system changes import status to ERROR, but adding jobProfileInfo at this point is not possible.

## Approach
Quick fix: Do not show UI records with jobProfileId and jobProfileName as null 

[MODSOURMAN-1186](https://folio-org.atlassian.net/browse/MODSOURMAN-1186)


RomanChernetskyi (approved):

Aliaksandr-Fedasiuk:
Another solution was found.

PR #907, opened 2024-05-17
fix(marc-fields-order): Don't group fields with same tags together

viacheslavkol:
### Purpose
Don't group fields with same tags together on fields reordering

### Related issues
[MODINV-1026](https://folio-org.atlassian.net/browse/MODINV-1026)
[MODSOURMAN-1191](https://folio-org.atlassian.net/browse/MODSOURMAN-1191)
[MODSOURCE-762](https://folio-org.atlassian.net/browse/MODSOURCE-762)

## Is this change testable? If not - why?
Will be tested in a scope of original SRS story


viacheslavpoliakov (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/afterprocessing/AdditionalFieldsUtil.java
```diff
@@ -175,60 +171,65 @@ public static boolean addFieldToMarcRecord(Record record, String field, char sub
...
-    var jsonNodesByTag = new LinkedHashMap<String, Queue<JsonNode>>();
-    for (JsonNode node : fieldsArrayNode) {
-      var tag = getTagFromNode(node);
-      jsonNodesByTag.putIfAbsent(tag, new LinkedList<>());
-      jsonNodesByTag.get(tag).add(node);
+  private static JsonNode removeAndGetNodeByTag(List<JsonNode> nodes, String tag) {
+    for (int i = 0; i < nodes.size(); i++) {
+      var nodeTag = getTagFromNode(nodes.get(i));
+      if (nodeTag.equals(tag)) {
+        return nodes.remove(i);
```
mukhiddin-yusuf, line 229: is it safe to remove in this way while doing traversal?

mukhiddin-yusuf: I think better to achieve this either with ```Iterator``` or ```List:: removeIf```

viacheslavkol: I think it's safe because we stop traversing on removal

mukhiddin-yusuf (approved):

PR #908, opened 2024-05-28
MODSOURMAN-1195 Save job execution progress in batches

okolawole-ebsco:
## Purpose
Save job execution progress in batches to improve performance and scalability.

## Approach
Instead of each record attempt to update job execution progress, a message is sent to an address over the Vert.x event bus. Upon successful placement of the message in the event bus, processing of the record continues.
On the consumer side, messages are group in batches by tenant ID and job execution ID. The group of messages are then reduced into a single job execution progress which is then saved into the database and overall job execution status is updated.
In local testing, update of job execution progress happens 10,000 times for a data import job of 10,000 records. Total execution time for all 10,000 database calls was **53 seconds**(mean execution time for each call was 5.3 milliseconds). After this optimization, there were only 352 database calls with a total execution time of **17 milliseconds**.

## Advantages
- Less database overhead due to less commits and less contention for the same database row.
- Less database connection pool thrashing since a connection is not required to update job execution progress for each record.
- Faster processing of records since processing is no longer waiting for update of job execution progress to continue.

## Disadvantages
- If for any reason the update of the reduced job execution fails, the records responsible for the update would have already been discarded. This is due to the asynchronous nature of this optimization.
- Very small data import jobs with minimal number of records can see total duration be extended by < 1000 milliseconds.


mod-source-record-manager-server/src/main/java/org/folio/services/RecordProcessedEventHandlingServiceImpl.java
```diff
@@ -103,67 +90,5 @@ private Future<JobExecution> updateJobStatusToError(String jobExecutionId, Okapi
       .withErrorStatus(StatusDto.ErrorStatus.FILE_PROCESSING_ERROR), params);
   }
 
-  private Future<Boolean> updateJobExecutionIfAllRecordsProcessed(String jobExecutionId, JobExecutionProgress progress, OkapiConnectionParams params) {
```
okolawole-ebsco, line 106: These lines were moved to the JobExecutionProgressVerticle file.

mod-source-record-manager-server/src/main/java/org/folio/services/progress/OptionalCodec.java
```diff
@@ -0,0 +1,36 @@
...
+
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.eventbus.MessageCodec;
+
+import java.util.Optional;
+
+/**
+ * Codec that won't cause any serialization/deserialization
+ */
+public class OptionalCodec implements MessageCodec<Optional<?>, Optional<?>> {
```
RuslanLavrov, line 11: could you say please why we need OptionalCodec codec? As far as understand a BatchableJobExecutionProgress object always is passed for event bus producer.

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/progress/OptionalCodec.java
okolawole-ebsco: Good catch! A previous implementation used this codec to notify the event bus producer that the batch action was completed. But it slowed data import jobs so i turned it to fire and forget. I have removed this codec since it is no longer needed.

PR #909, opened 2024-05-29
MODSOURCE-733: raml-storage dependency updated.

VRohach:
## Purpose
_Describe the purpose of the pull request. Include background information if necessary._

## Approach

## Checklist
- [ ] I have updated NEWS.md.


mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
```diff
@@ -12,13 +12,7 @@
...
 import org.folio.dataimport.util.OkapiConnectionParams;
 import org.folio.kafka.exception.DuplicateEventException;
-import org.folio.rest.jaxrs.model.EntityType;
-import org.folio.rest.jaxrs.model.InitialRecord;
-import org.folio.rest.jaxrs.model.JobExecution;
-import org.folio.rest.jaxrs.model.JobExecutionSourceChunk;
-import org.folio.rest.jaxrs.model.ProfileSnapshotWrapper;
-import org.folio.rest.jaxrs.model.RawRecordsDto;
-import org.folio.rest.jaxrs.model.StatusDto;
+import org.folio.rest.jaxrs.model.*;
```
Maksat-Galymzhan, line 15: Please, use single class import instead of wild card import

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
```diff
@@ -27,7 +21,7 @@
 import java.util.List;
 import java.util.Map;
 
-import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.*;
+import static org.folio.rest.jaxrs.model.ProfileType.*;
```
Maksat-Galymzhan, line 24: the same

Maksat-Galymzhan (approved):

RomanChernetskyi (approved):

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
VRohach: @Maksat-Galymzhan  Thank you! It seems like after updating IntelliJIdea this setting was removed. Fixed.

mod-source-record-manager-server/src/main/java/org/folio/services/AbstractChunkProcessingService.java
VRohach: Fixed.

PR #913, opened 2024-07-24
MODSOURMAN-1028 Save Journal Records In Batches

okolawole-ebsco:
## Purpose
Save journal records in batches to reduce resource consumption.

## Approach
- Instead of one kafka consumer for every topic needed for journaling, use one kafka consumer for all needed topics. This will reduce the number of connections to the kafka clusters and less CPU/Memory on the module instance
- Instead of committing upon every message (via folio-kafka-wrapper's KafkaConsumerWrapper), messages are committed in batches. This saves the kafka cluster some work.
- Instead of tracking processed messages in the database table, the single kafka consumer is now idempotent. The properties of a journal record is used to derive a UUID for the journal record; representing its identifier. This means that if the same message is consumed numerous times, it will resolve to the same journal record with the same identifier. Insert of the same record will have its error suppressed(using `ON CONFLICT` clause of postgres)
- Journal records are grouped together and committed in batches. This saves database overhead with committing every row.


Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -0,0 +1,425 @@
...
+                .map(this::setDeterministicIdentifer)
+                .toList();
+              // If no records or tenant ID is missing, complete without action
+              if (journalRecords.isEmpty() || groupedRecords.getKey().isEmpty()) return Completable.complete();
+
+              LOGGER.info("saveJournalRecords:: Saving {} journal record(s) for tenantId={}", journalRecords.size(), groupedRecords.getKey().get());
+              // Save the batch of journal records and handle the response
+              return AsyncResultCompletable.toCompletable(completionHandler -> batchJournalService.saveBatchWithResponse(
+                journalRecords,
+                groupedRecords.getKey().get(),
```
RuslanLavrov, line 311: found that looks like we must not pass tenantId from the journal record to the method for saving journal records. Because journalRecord.tenantId field is used to keep id of tenant where entity/record is located over which action has been applied. This field is mostly used to keep central tenant id of the instance/record action on which action has been applied during data import initiated on the member tenant to ensure appearance of shared instance marc record and proper work of link to the updated shared instance on the data import logs pages. If we pass the tenantId from the journal record instead from the headers during a shared instance update by import from a member tenant, it will result in some journal records being saved to the central tenant DB schema instead of the member tenant schema, causing to the loss of logs data for the import job on the member tenant.
In the case of shared instance update by import from a member tenant, the tenantId in the headers will contain the member tenant id, and the journal record for the shared instance update will contain the tenantId of the central tenant.
Maybe we can add tenantId field to the BatchableJournalRecord class to keep there always tenant id from the headers, and then use tenantId from the BatchableJournalRecord to group journal records by tenant id before saving.
In this way, when passing groupedRecords key as a tenantId for journal record saving, we will always pass the tenantId from the headers.

okolawole-ebsco: I have updated the code, please check it out!

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
RuslanLavrov: thank you!

PR #926, opened 2024-09-09
MODSOURMAN-1212: Update MARC bib-instance default mapping for subject source and subject type.

VRohach:
## Purpose
In order to support further subject heading granularity, we need to update the default MARC-Instance map to populate the new fields for Subject source and Subject type.

## Approach
Default mapping rules improved to be able to provide mapping for Subject Source and Subject Type based specific conditions from JIRA-ticket.
**DEFAULT MAPPING RULES EXTENDED**

JIRA: https://folio-org.atlassian.net/browse/MODSOURMAN-1212


VRohach:
THIS PR SHOULD BE MERGED ONLY AFTER https://github.com/folio-org/data-import-processing-core/pull/352

KaterynaSenchenko (approved):

JavokhirAbdullayev (approved):

PR #936, opened 2024-10-08
MODSOURMAN-1228: Update default mapping for Date type, Date 1, and Date 2 fields

VRohach:
## Purpose
The main purpose is to update default mapping for Date type, Date 1, and Date 2 fields

## Approach
So there should be extended our default mapper. Previouslym we did NOT map simple JsonObjects, just String or Arrays. In our case, the "Dates" - is an object, and our mapper can't work with this type of Instance's field. So there was added a new property to rules named "createSingleObject". And if mapper find this rules - it will handle this rule via specific logic, building rules into 1 JsonObject. 

1. Mapping parameters extended via new Date data.
2. Default Mapping rules improved for 008 field.
3. Tests improved.
4. New permissions added.
5. Documentation for mapping rules extended via new property.
## Learning
JIRA: https://folio-org.atlassian.net/browse/MODSOURMAN-1228

**NOTE**: This PR should be merged **ONLY** with these 2: https://github.com/folio-org/data-import-processing-core/pull/367 AND https://github.com/folio-org/mod-data-import/pull/340

### DIFF
--- a/mod-source-record-manager-server/src/main/resources/rules/marc_bib_rules.json
+++ b/mod-source-record-manager-server/src/main/resources/rules/marc_bib_rules.json
@@ -77,7 +77,6 @@
       "createSingleObject": true,
       "rules": [
         {
-          "description": "",
           "conditions": [
             {
               "type": "set_date_type_id"
--- a/mod-source-record-manager-server/src/main/resources/rules/marc_bib_rules.json
+++ b/mod-source-record-manager-server/src/main/resources/rules/marc_bib_rules.json
@@ -69,6 +69,60 @@
           ]
         }
       ]
+    },
+    {
+      "target": "dates.dateTypeId",
+      "description": "Date type ID",
+      "subfield": [],
+      "createSingleObject": true,
+      "rules": [
+        {
+          "description": "",
+          "conditions": [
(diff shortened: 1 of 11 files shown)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/resources/rules/marc_bib_rules.json
```diff
@@ -69,6 +69,60 @@
...
       ]
+    },
+    {
+      "target": "dates.dateTypeId",
+      "description": "Date type ID",
+      "subfield": [],
+      "createSingleObject": true,
+      "rules": [
+        {
+          "description": "",
```
RuslanLavrov, line 80: can we not specify this empty field?

VRohach: @RuslanLavrov  Oh I will just remove this field, Thank you!

RuslanLavrov: got it, thank you!

PR #937, opened 2024-10-11
MODSOURMAN-1232 Add the option to exclude job profile names to GET "/metadata-provider/jobExecutions" endpoint

JavokhirAbdullayev:
## Purpose
[MODSOURMAN-1232](https://folio-org.atlassian.net/browse/MODSOURMAN-1232) Add the option to exclude job profile names to GET "/metadata-provider/jobExecutions" endpoint

## Approach
add new query param to endpoint

## Checklist
- [x] I have updated NEWS.md.


mod-source-record-manager-server/src/test/java/org/folio/rest/impl/metadataProvider/MetadataProviderJobExecutionAPITest.java
```diff
@@ -547,6 +547,21 @@ public void shouldReturnFilteredCollectionByHrIdOnGet() {
...
   }
 
+  @Test
+  public void shouldReturnFilteredCollectionByExcludeJobProfileNameOnGet() {
+    constructAndPostInitJobExecutionRqDto(5);
+
+    RestAssured.given()
+      .spec(spec)
+      .when()
+      .queryParam("excludeJobProfileNames", "importBib2.bib")
```
KaterynaSenchenko, line 557: please add test to exclude several jobs that have similar names satisfying filter with wildcard

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionFilter.java
```diff
@@ -121,16 +129,19 @@ public String buildCriteria() {
...
-        addCondition(conditionBuilder, buildCaseSensitiveLikeCondition(HRID_FIELD, hrIdPattern));
+        addCondition(conditionBuilder, buildCaseSensitiveLikeCondition(HRID_FIELD, hrIdPattern, false));
       }
       if (isNotEmpty(fileNamePattern)) {
-        addCondition(conditionBuilder, buildCaseInsensitiveLikeCondition(FILE_NAME_FIELD, fileNamePattern));
+        addCondition(conditionBuilder, buildCaseInsensitiveLikeCondition(FILE_NAME_FIELD, fileNamePattern, false));
       }
     }
+    if (isNotEmpty(excludeJobProfileNames)) {
+      addCondition(conditionBuilder, buildCaseInsensitiveLikeCondition(FILE_NAME_FIELD, excludeJobProfileNames, true));
```
RuslanLavrov, line 143:
```suggestion
      addCondition(conditionBuilder, buildCaseInsensitiveLikeCondition(JOB_PROFILE_NAME_FIELD, excludeJobProfileNames, true));
```
please, update this line with JOB_PROFILE_NAME_FIELD because here we must filter out jobs by job profile name instead of file name

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JobExecutionFilter.java
RuslanLavrov: thanks!

PR #949, opened 2024-11-15
MODSOURMAN-1249: "SRS MARC" and "Authority" tabs are empty in data import of MARC authority log

Aliaksandr-Fedasiuk:
## Purpose
DI log is not opened on UI and data is not written to the database

## Approach
Added DI_MARC_FOR_UPDATE_RECEIVED log message handling

## Checklist
- [x] I have updated NEWS.md.


KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -158,6 +159,7 @@ public List<String> getEvents() {
       DI_SRS_MARC_AUTHORITY_RECORD_NOT_MATCHED.value(),
       DI_SRS_MARC_HOLDINGS_RECORD_NOT_MATCHED.value(),
       DI_SRS_MARC_HOLDINGS_RECORD_MATCHED.value(),
+      DI_MARC_FOR_UPDATE_RECEIVED.value(),
```
RuslanLavrov, line 162: please, don't add this event because we should not subscribe to it in order not to save redundant/incorrect data to log

Aliaksandr-Fedasiuk: Fixed. Thanks!

RuslanLavrov (approved):

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
RuslanLavrov: thank!

PR #953, opened 2025-01-13
MODDATAIMP-1135 Log entries are empty for completed jobs

JavokhirAbdullayev:
## Purpose
[MODDATAIMP-1135](https://folio-org.atlassian.net/browse/MODDATAIMP-1135) Log entries are empty for completed jobs

## Approach
 * save journal recs by job id
* group journal records by job execution id to reduce chances of deadlock in the database. 
* Also modified DataImportJournalBatchConsumerVerticle to be resilient to errors and continue processing without restarting the service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,10 @@ which are created during tenant initialization, the values of which can be custo
 and `DI_RAW_RECORDS_CHUNK_PARSED_PARTITIONS` env variables respectively.
 Default value - `1`.
 
+## Environment variables
+`MAX_NUM_EVENTS`:`100` - set max num events to consume
+
+
 #### Note:
 From v 3.1.3 there is a new property which defines limit for retrieving data to fill mapping parameters for the data-import mechanism: **"srm.mapping.parameters.settings.limit:1000"**
 
--- a/README.md
+++ b/README.md
@@ -132,7 +132,6 @@ Default value - `1`.
 ## Environment variables
 `MAX_NUM_EVENTS`:`100` - set max num events to consume
 
-
 #### Note:
 From v 3.1.3 there is a new property which defines limit for retrieving data to fill mapping parameters for the data-import mechanism: **"srm.mapping.parameters.settings.limit:1000"**
 
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
@@ -137,8 +137,6 @@ import static org.folio.rest.persist.PostgresClient.convertToPsqlStandard;
 public class JournalRecordDaoImpl implements JournalRecordDao {
 
   private static final Logger LOGGER = LogManager.getLogger();
-  private static final int MAX_RETRIES = 3;
-  private static final long INITIAL_BACKOFF_MS = 50;
   public static final String SOURCE_RECORD_ENTITY_TYPE = "source_record_entity_type";
   public static final String ORDER_ENTITY_ID = "order_entity_id";
   public static final String INCOMING_RECORD_ID = "incoming_record_id";
--- a/mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
@@ -137,6 +137,8 @@ import static org.folio.rest.persist.PostgresClient.convertToPsqlStandard;
 public class JournalRecordDaoImpl implements JournalRecordDao {
 
   private static final Logger LOGGER = LogManager.getLogger();
+  private static final int MAX_RETRIES = 3;
+  private static final long INITIAL_BACKOFF_MS = 50;
   public static final String SOURCE_RECORD_ENTITY_TYPE = "source_record_entity_type";
   public static final String ORDER_ENTITY_ID = "order_entity_id";
   public static final String INCOMING_RECORD_ID = "incoming_record_id";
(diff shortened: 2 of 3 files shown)


Aliaksandr-Fedasiuk (approved):

KaterynaSenchenko (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -180,18 +182,80 @@ public Future<String> save(JournalRecord journalRecord, String tenantId) {
...
+      finalPromise.fail(err);
+    });
+  }
+
+  private void executeGroup(String query,
+                            List<Tuple> tuples,
+                            String tenantId,
+                            Promise<List<RowSet<Row>>> promise) {
+    pgClientFactory.createInstance(tenantId)
+      .execute(query, tuples, ar -> {
```
JavokhirAbdullayev, line 251: if we create PgInstance for each group within batch batch, wouldn't it lead to a memory leak?

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -238,11 +268,13 @@ private Flowable<Pair<Optional<Bundle>, Collection<BatchableJournalRecord>>> lis
...
+      .onBackpressureBuffer(10_000,
+        () -> LOGGER.error("listenKafkaEvents:: Backpressure buffer full"),
+        BackpressureOverflowStrategy.DROP_OLDEST)
       .map(consumerRecord -> {
         try {
           Map<String, String> map = kafkaHeadersToMap(consumerRecord.headers());
-          OkapiConnectionParams okapiConnectionParams = OkapiConnectionParams.createSystemUserConnectionParams(
-            map, vertx.getDelegate());
+          OkapiConnectionParams okapiConnectionParams = new OkapiConnectionParams(map, vertx.getDelegate());
           String recordId = okapiConnectionParams.getHeaders().get(RECORD_ID_HEADER);
```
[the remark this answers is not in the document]
JavokhirAbdullayev: I see, it does not matter here since we are not publishing any events, only consuming events

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -238,11 +268,13 @@ private Flowable<Pair<Optional<Bundle>, Collection<BatchableJournalRecord>>> lis
       throw new IllegalStateException("KafkaConsumer not initialized");
     }
     return kafkaConsumer.toFlowable()
+      .onBackpressureBuffer(10_000,
+        () -> LOGGER.error("listenKafkaEvents:: Backpressure buffer full"),
+        BackpressureOverflowStrategy.DROP_OLDEST)
```
JavokhirAbdullayev, line 273: just to clarify, if we discard old events, they won't be lost, right? as their offsets were not committed to broker, and broker will still send those discard events later on?

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
JavokhirAbdullayev: i see, this method does not create client it creates only if PostgresClient does not exist

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -137,6 +137,8 @@
 public class JournalRecordDaoImpl implements JournalRecordDao {
 
   private static final Logger LOGGER = LogManager.getLogger();
+  private static final int MAX_RETRIES = 3;
+  private static final long INITIAL_BACKOFF_MS = 50;
```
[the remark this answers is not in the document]
okolawole-ebsco: Ah, lets remove that. I had an implementation for exponential backoff during retry but i felt it was too complex. You can remove these lines.

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
okolawole-ebsco: They would be lost if new offsets are committed to the database.

I read the code for the FlowableReadStream in vert.x and it looks like the kafka consumer is automatically paused when there is backpressure. So please remove this change and lets rely on the buffer from vertx. https://github.com/vert-x3/vertx-rx/blob/96287695ac0a58ad01bf9331088ca69242d87564/rx-java3-gen/src/main/java/io/vertx/rxjava3/impl/FlowableReadStream.java#L72

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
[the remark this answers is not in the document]
JavokhirAbdullayev: yeah, i guess we can these features in a separate ticket, like we did for mod-srs

README.md
```diff
@@ -129,6 +129,9 @@ which are created during tenant initialization, the values of which can be custo
 and `DI_RAW_RECORDS_CHUNK_PARSED_PARTITIONS` env variables respectively.
 Default value - `1`.
 
+## Environment variables
+`MAX_NUM_EVENTS`:`100` - set max num events to consume
```
okolawole-ebsco, line 133: For future reference, we should name the env var better so that it communicates its domain at batch journaling rather than the entire application.

JavokhirAbdullayev: thank you! will update in later PR

PR #955, opened 2025-01-16
MODSOURMAN-1268 Log entries are empty for completed jobs

JavokhirAbdullayev:
## Purpose
increase session timeouts and heartbeats for consumer

## Approach

## Checklist
- [ ] I have updated NEWS.md.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
@@ -223,9 +223,9 @@ public class DataImportJournalBatchConsumerVerticle extends AbstractVerticle {
     Map<String, String> consumerProps = kafkaConfigWithDeserializer.getConsumerProps();
     // this is set so that this consumer can start where the non-batch consumer left off, when no previous offset is found.
     consumerProps.put(KafkaConfig.KAFKA_CONSUMER_AUTO_OFFSET_RESET_CONFIG, "latest");
-
-    consumerProps.put(KafkaConfig.KAFKA_CONSUMER_MAX_POLL_INTERVAL_MS_CONFIG, "600000");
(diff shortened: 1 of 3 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -223,6 +224,15 @@ private Completable initializeKafkaConsumer() {
     Map<String, String> consumerProps = kafkaConfigWithDeserializer.getConsumerProps();
     // this is set so that this consumer can start where the non-batch consumer left off, when no previous offset is found.
     consumerProps.put(KafkaConfig.KAFKA_CONSUMER_AUTO_OFFSET_RESET_CONFIG, "latest");
+
+    consumerProps.put(KafkaConfig.KAFKA_CONSUMER_MAX_POLL_INTERVAL_MS_CONFIG, "900000");
```
RuslanLavrov, line 228: please, specify these constants:
ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG
ConsumerConfig.MAX_POLL_RECORDS_CONFIG
ConsumerConfig.AUTO_OFFSET_RESET_CONFIG
 
instead of these
KafkaConfig.KAFKA_CONSUMER_MAX_POLL_INTERVAL_MS_CONFIG
KafkaConfig.KAFKA_CONSUMER_MAX_POLL_RECORDS_CONFIG
KafkaConfig.KAFKA_CONSUMER_AUTO_OFFSET_RESET_CONFIG

so that the parameters values can take effect

JavokhirAbdullayev: but with params from KafkaConfig it also works, I will try with default max pool interval

RuslanLavrov: yes, but if we need to set our values for the parameters then the ConsumerConfig constants should be used

PR #957, opened 2025-01-22
Remove polling when RebalanceInProgressException occured during consumption of events 

JavokhirAbdullayev:
## Purpose
in the previous approach, we were waiting, polling, and then committing offsets if a  rebalance exception occurred in the commitOffsets logic 
now, we wait then retry commit without polling for new records
also other error are handled in similar fashion

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
@@ -422,8 +422,7 @@ public class DataImportJournalBatchConsumerVerticle extends AbstractVerticle {
       .onErrorResumeNext(error -> {
         if (error instanceof RebalanceInProgressException) {
           LOGGER.warn("Rebalance in progress. Waiting for the re-balance to complete...");
-          return Completable.timer(100, TimeUnit.MILLISECONDS)
-            .andThen(commitOffset(offsets));
+          return Completable.timer(1, TimeUnit.MILLISECONDS);
         }
         LOGGER.warn("Error committing offsets: {}. Retrying in 1 second...", error.getMessage());
         return Completable.timer(1, TimeUnit.SECONDS)
--- a/mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
@@ -422,7 +422,7 @@ public class DataImportJournalBatchConsumerVerticle extends AbstractVerticle {
       .onErrorResumeNext(error -> {
         if (error instanceof RebalanceInProgressException) {
           LOGGER.warn("Rebalance in progress. Waiting for the re-balance to complete...");
-          return Completable.timer(1, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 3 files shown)


mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -421,11 +421,13 @@ private Completable commitOffset(Map<TopicPartition, OffsetAndMetadata> offsets)
       )
       .onErrorResumeNext(error -> {
         if (error instanceof RebalanceInProgressException) {
-          LOGGER.warn("Rebalance in progress. Retrying offset commit...");
-          return kafkaConsumer.rxPoll(Duration.ofMillis(100))
-            .flatMapCompletable(records -> commitOffset(offsets));
```
JavokhirAbdullayev, line 426: but here we were just polling tp finish the rebalance and committing the same offsets? 

`2025-01-18T15:28:15.450588353Z org.apache.kafka.common.errors.RebalanceInProgressException: Offset commit cannot be completed since the consumer is undergoing a rebalance for auto partition assignment. You can try completing the rebalance by calling poll() and then retry the operation.`

okolawole-ebsco: Remove commit of offsets as discussed if it is generating an error.

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -421,11 +421,13 @@ private Completable commitOffset(Map<TopicPartition, OffsetAndMetadata> offsets)
...
+          LOGGER.warn("Rebalance in progress. Waiting for the re-balance to complete...");
+          return Completable.timer(100, TimeUnit.MILLISECONDS)
```
okolawole-ebsco, line 425: I think 100 milliseconds is too short of a time. Give it a second at least.

mod-source-record-manager-server/src/main/java/org/folio/verticle/DataImportJournalBatchConsumerVerticle.java
```diff
@@ -421,11 +421,12 @@ private Completable commitOffset(Map<TopicPartition, OffsetAndMetadata> offsets)
       )
       .onErrorResumeNext(error -> {
         if (error instanceof RebalanceInProgressException) {
-          LOGGER.warn("Rebalance in progress. Retrying offset commit...");
-          return kafkaConsumer.rxPoll(Duration.ofMillis(100))
-            .flatMapCompletable(records -> commitOffset(offsets));
+          LOGGER.warn("Rebalance in progress. Waiting for the re-balance to complete...");
+          return Completable.timer(1, TimeUnit.MILLISECONDS);
```
okolawole-ebsco, line 425: This is one millisecond instead of one second?

JavokhirAbdullayev: fixed :)

okolawole-ebsco (approved):

KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

PR #965, opened 2025-02-17
MODSOURMAN-1281: 500 Error Returned when Posting Records Chunk with Existing ID Value

VRohach:
## Purpose
The main purpose is to change from 500 status code to 400 for the `/change-manager/jobExecutions/{job_id}/records `endpoint if the same id was received.

## Approach
Additional condition added.

JIRA: https://folio-org.atlassian.net/browse/MODSOURMAN-1281


mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -204,7 +206,14 @@ public void postChangeManagerJobExecutionsRecordsById(String id, boolean acceptI
...
+            if (ex instanceof DuplicateEventException) {
+              LOGGER.warn("postChangeManagerJobExecutionsRecordsById:: Failed to process chunk of RawRecords with JobExecution id {}: {}", id, ex.getMessage());
+              return ExceptionHelper.mapExceptionToResponse(new BadRequestException(ex.getMessage()));
+            } else {
+              return ExceptionHelper.mapExceptionToResponse(ex);
+            }
+          })
           .onComplete(asyncResultHandler);
       } catch (Exception e) {
         LOGGER.warn(getMessage("postChangeManagerJobExecutionsRecordsById:: Failed to process chunk of RawRecords with JobExecution id {}", e, id));
```
Aliaksandr-Fedasiuk, line 219: I think that the order for "e" and "id" is incorrect.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
```diff
@@ -204,7 +206,14 @@ public void postChangeManagerJobExecutionsRecordsById(String id, boolean acceptI
         eventDrivenChunkProcessingService.processChunk(entity, id, acceptInstanceId, params)
           .map(processed -> PostChangeManagerJobExecutionsRecordsByIdResponse.respond204())
           .map(Response.class::cast)
-          .otherwise(ExceptionHelper::mapExceptionToResponse)
+          .otherwise(ex -> {
+            if (ex instanceof DuplicateEventException) {
+              LOGGER.warn("postChangeManagerJobExecutionsRecordsById:: Failed to process chunk of RawRecords with JobExecution id {}: {}", id, ex.getMessage());
```
Aliaksandr-Fedasiuk, line 211: Please add jobExecutionId and recordId to the logs where possible.

Aliaksandr-Fedasiuk (changes requested):

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
VRohach: @Aliaksandr-Fedasiuk No, you are not correct. Please look at how arguments inside the method "getMessage()" are provided.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
VRohach: @Aliaksandr-Fedasiuk Thank you! Added "entity.getId()" (RawRecordsId) to this message.

mod-source-record-manager-server/src/main/java/org/folio/rest/impl/ChangeManagerImpl.java
Aliaksandr-Fedasiuk: Hmm. Thanks, but are we sure we need to use the getMessage method?

VRohach: @Aliaksandr-Fedasiuk, I hope this is needed for formatting logs with an exception. This is used in all endpoints in this module. Let's leave it here.

Aliaksandr-Fedasiuk: Thanks

Aliaksandr-Fedasiuk (approved):

JavokhirAbdullayev (approved):

NEWS.md
```diff
@@ -2,6 +2,7 @@
 * [MODSOURMAN-1249](https://folio-org.atlassian.net/browse/MODSOURMAN-1249) Added DI_MARC_FOR_UPDATE_RECEIVED log message handling
 * [MODSOURMAN-1266](https://folio-org.atlassian.net/browse/MODSOURMAN-1266) Event DI_JOB_COMPLETED is not being sent upon the completion of the data import process on snapshot, snapshot-2
 * [MODSOURMAN-1276](https://folio-org.atlassian.net/browse/MODSOURMAN-1276) Additional 147/447/547 to authority mappings
+* [MODSOURMAN-1281](https://folio-org.atlassian.net/browse/MODSOURMAN-1281) Event DI_JOB_COMPLETED is not being sent upon the completion of the data import process on snapshot, snapshot-2
```
KaterynaSenchenko, line 5: Please correct the news entry

VRohach: @KaterynaSenchenko  Thanks for pointing this.

PR #966, opened 2025-02-21
MODDICORE-440 Mode of Issuance not being set with MARC to Instance Mapping

JavokhirAbdullayev:
## Purpose
MODDICORE-440 Mode of Issuance not being set with MARC to Instance Mapping

## Approach
* added new rule for LDR


dmytrokrutii (approved):

JavokhirAbdullayev:
should be merged after https://github.com/folio-org/data-import-processing-core/pull/389
that's why tests are failing

RuslanLavrov (approved):

PR #967, opened 2025-02-25
MODSOURMAN-1294 Data-import. Slow DB query [SELECT insert_journal_records($1::jsonb[])]

JavokhirAbdullayev:
## Purpose
[MODSOURMAN-1294](https://folio-org.atlassian.net/browse/MODSOURMAN-1294)Data-import. Slow DB query [SELECT insert_journal_records($1::jsonb[])]

## Approach
* add retry logic with exponential back-off when deadlock arises


mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -183,36 +187,47 @@ public Future<Void> saveBatch(Collection<JournalRecord> journalRecords, String t
...
+      List<Tuple> tupleList = journalRecords.stream()
+        .map(this::prepareInsertQueryParameters)
+        .toList();
+      String query = format(INSERT_SQL, convertToPsqlStandard(tenantId), JOURNAL_RECORDS_TABLE);
+      LOGGER.trace("saveBatch:: query = {}; tuples = {}", query, tupleList);
+
+      return executeWithRetry(query, tupleList, tenantId, MAX_RETRIES, INITIAL_RETRY_DELAY_MS)
+        .mapEmpty();
     } catch (Exception e) {
       LOGGER.warn("saveBatch:: Error saving journal records", e);
```
Aliaksandr-Fedasiuk, line 199: I would use the ERROR logging level here.

Aliaksandr-Fedasiuk (approved):

mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
JavokhirAbdullayev: in DataImportJournalBatchConsumerVerticle we are already logging it with error level.

okolawole-ebsco (approved):

KaterynaSenchenko (approved):

PR #972, opened 2025-03-11
MODSOURMAN-1287: DI completion event is produced more than once

Aliaksandr-Fedasiuk:
## Purpose
_Describe the purpose of the pull request. Include background information if necessary._

## Approach

## Checklist
- [ ] I have updated NEWS.md.

### DIFF
--- a/mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
+++ b/mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
@@ -22,6 +22,7 @@ import org.folio.services.exceptions.RecordsPublishingException;
 import static org.folio.services.util.RecordConversionUtil.RECORDS;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 public final class EventHandlingUtil {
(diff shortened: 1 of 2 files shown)


mod-source-record-manager-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -44,33 +44,40 @@ private EventHandlingUtil() {
...
+      LOGGER.info("logSendingSucceeded:: Event with type: {} for jobExecutionId: {} and chunkId: {} was sent to kafka",
+        eventType, jobExecutionId, chunkId);
     } else {
-      LOGGER.info("logSendingSucceeded:: Event with type: {} and recordId: {} was sent to kafka", eventType, recordId);
+      LOGGER.info("logSendingSucceeded:: Event with type: {} for jobExecutionId: {} and recordId: {} was sent to kafka",
+        eventType, jobExecutionId, recordId);
     }
   }
 
+  public static String extractJobExecutionId(String eventPayload) {
```
VRohach, line 76: @Aliaksandr-Fedasiuk  Maybe it would be better to make it private and non-static?
Or if we need it static, move it to the specific Util class.
What do you think?

Aliaksandr-Fedasiuk: Thanks, this is the unnecessary method. Removed.

VRohach (approved):

RuslanLavrov (approved):

PR #992, opened 2025-09-23
MODSOURMAN-1355: Add consistent record ordering to prevent deadlocks in journal record insertions

Aliaksandr-Fedasiuk:
## Purpose
When writing journal records in batches to the log, messages about deadlocks appeared, which were resolved with the help of the retry mechanism.

## Approach
Sorting records will help reduce the number of deadlock cases.

## Learn
[MODSOURMAN-1355](https://folio-org.atlassian.net/browse/MODSOURMAN-1355)


mod-source-record-manager-server/src/main/java/org/folio/dao/JournalRecordDaoImpl.java
```diff
@@ -625,15 +632,25 @@ private static RecordProcessingLogDtoCollection processMultipleHoldingsAndItemsI
   }
 
   private static boolean ifNeedToMerge(List<RecordProcessingLogDto> entries) {
```
okolawole-ebsco, line 634: why does this function need to change?

PR #993, opened 2025-09-25
fix(custom-migration) Fix: Mapping rules migrations were not idempotent

SvitlanaKovalova1:
[MODSOURMAN-1357](https://folio-org.atlassian.net/browse/MODSOURMAN-1357) Mapping rules migrations are not idempotent
## Purpose
- Mapping rules migrations should be idempotent.
- Each migration should be executed only once per module version in which it was introduced.

## Approach
Save custom migration history in the database to avoid duplicate execution.

## Is this change testable? yes
Applied list of custom migrations:
image

## Checklist
- [x] I have updated NEWS.md.
- [x] I have added javadocs to new methods.


mod-source-record-manager-server/src/main/resources/templates/db_scripts/create_rule_migration_change_log_table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS rule_migration_change_log (
```
psmagin, line 1: You should also populate the table with migrations that are expected to already be executed before this PR.

psmagin (changes requested):

